PR #3, opened 2017-06-21
Add ContravariantDerivation for cats.Show to examples

ruippeixotog:
Adds the implicits required to derive `cats.Show` type classes to the examples project and also adds a runnable test in the tests project.


propensive:
Thank you for your work on this! I've just finished cleaning it up and merging it!

PR #5, opened 2017-06-22
Use case clauses and switch statement instead of is/asInstanceOf

propensive:


core/src/main/scala/magnolia.scala
```diff
@@ -154,18 +154,15 @@ class Macros(val c: whitebox.Context) {
...
-              q"""
-                if(sourceParameter.isInstanceOf[$componentType])
-                  $impl.call($derivedImplicit, sourceParameter.asInstanceOf[$componentType])
-                else $aggregated"""
+            val caseClauses = parts.map { case (subtype, component) =>
+              cq"(value: $subtype) => $impl.call($component, value)"
             }
+
+            q"""(sourceParameter: @_root_.scala.annotation.switch) match {
+              case ..$caseClauses
```
fommil: where are you generating the `Int` that sits on each instance? btw, make sure it's not a `val` or it'll add to the heap cost.

andyscott: 👏 👏 👏

propensive: I'm not. I thought that was automatic for sealed traits? Shouldn't it be a compile error if you try to `@switch`, and it doesn't have it?

fommil: Nope, it needs to be added... that's where my thing comes in. I bet the compiler just ignores the switch request.

PR #17, opened 2017-10-30
bump versions, make Show stalactite compatible

fommil:
I can't go any further because it depends on unreleased versions...

I also seem to have unearthered a bug in your macro. Or my macro. Or macro paradise.


propensive:
Closing this as the build file has changed too much since then. I think that from the current configuration, the changes required to support `stalactite` would be minimal... (just the annotation, maybe)

PR #26, opened 2017-11-10
Remove Scala 2.12 deprecation compiler warning

### LINKED ISSUE #19: Remove Scala 2.12 deprecation compiler warning
When compiling the Magnolia macro, the compiler warns that `companionSymbol` is deprecated. This code should be replaced with something which works on both Scala 2.11 and 2.12.

mielientiev:
it should fix #19

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -44,7 +44,7 @@ lazy val buildSettings = Seq(
   ),
   scmInfo := Some(
     ScmInfo(url("https://github.com/propensive/magnolia"),
-            "scm:git:git@github.com:propensive/magnolia.git")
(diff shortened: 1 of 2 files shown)


propensive:
Thank you! I thought it was more complicated than that when I tried it, though. Did you run the tests?
```
sbt> tests/run
```
should do it.

mielientiev:
@propensive you right for `isCaseObject` case it doesn't work... 
At least the second `companionSymbol` fixes deprecation :D

propensive:
I found some code in Rapture where I had worked around the same problem, so I've fixed it in a separate PR, so I'll close this one now.

PR #30, opened 2017-11-10
Add minimal testcase for testing functor derivation.

LukaJCB:
As requested :)

The Functor type class should be something like 
```scala
trait Functor[F[_]] {
  def map[A, B](fa: F[A])(f: A => B): F[B]
}
```

and the manual instance would look something like this:
```scala

implicit val functorTree: Functor[Tree] = new Functor[Tree] {
  def map[A, B](fa: Tree[A])(f: A => B): Tree[B] = fa match {
    case Leaf(a) => Leaf(f(a))
    case Branch(left, right) => Branch(map(left)(f), map(right)(f))
  }
}
```

### DIFF
--- a/tests/src/main/scala/tests.scala
+++ b/tests/src/main/scala/tests.scala
@@ -129,6 +129,10 @@ object Tests extends TestApp {
       implicitly[Default[Item]].default
     }.assert(_ == Item("", 1, 0))
 
+    test("functor works as expected") {
+      implicitly[Functor[Tree]].map(Branch(Leaf("Foo"), Leaf("Hello")))(_.length)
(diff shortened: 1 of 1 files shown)


propensive:
Ok, let's give this a go!

propensive:
I'm not sure I'm making much progress with this. I think it will require some more changes to Magnolia, and I'm not currently sure what they are...

propensive:
I've postponed trying to find a solution to this until after the 1.0 release.

propensive:
I wonder whether we might be closer to implementing this once I've finished implementing #46. I think it's likely to be a step closer, so I'll keep this PR open until that's merged.

LukaJCB:
I'm going to close this for now :)

PR #39, opened 2017-11-13
Restore ability to derive typeclasses for case classes inside other classes

shadaj:
In the process of upgrading [Slinky](https://github.com/shadaj/slinky) to the latest version of Magnolia, I noticed that the macro now throws an exception when attempting to derive a typeclass for a case class within another class. This PR restores the logic for getting all case class parameters to the one used in Magnolia 0.2.0, which fixes the errors in Slinky's [unit tests](https://github.com/shadaj/slinky/blob/master/tests/src/test/scala/me/shadaj/slinky/core/ReaderWriterTest.scala).


propensive:
Thank you for this! I'm sorry I didn't see it sooner... would you mind rebasing it? Adding a test case to `tests.scala` would also be great (and I can use it verify the bad behaviour with the current version, too)!

shadaj:
Just pushed up a rebase and unit tests! I couldn't figure out how to get default parameter values of an inner case class, so I ended up only supporting derivation of inner case classes when they have no default parameter values.

core/shared/src/main/scala/magnolia.scala
```diff
@@ -278,17 +278,26 @@ object Magnolia {
...
-              if (p.isParamWithDefault) {
-                val method = TermName("apply$default$" + (idx + 1))
-                q"$scalaPkg.Some(${genericType.typeSymbol.companion.asTerm}.$method)"
-              } else q"$scalaPkg.None"
+          val constructorParams = genericType.decls.collect {
+            case a: MethodSymbol if a.isConstructor => a
+          }.head.paramLists.head.map(_.asTerm)
+          val noDefaults = constructorParams.forall(!_.isParamWithDefault)
+
+          if (noDefaults) {
```
propensive: What is the purpose of the special case when none of the parameters have default values, rather than just letting the `else` block deal with each one in turn?

tests/src/main/scala/tests.scala
```diff
@@ -176,5 +176,13 @@ object Tests extends TestApp {
     test("serialize a value class") {
       Show.gen[Length].show(new Length(100))
     }.assert(_ == "100")
+
```
propensive: This test does not actually get run when you call `sbt tests/run`! The reason is that the test library only runs tests which are in the execution path, and as `ParentClass` is not instantiated anywhere, it doesn't get run! (NB. This is not explained *anywhere*, so can be completely forgiven! ;) The test ran successfully when I tried it.)

propensive: Actually, I just tested a refactored version of this test on the current release, `0.6.1`, and it worked fine, because I was misunderstanding exactly what you were fixing. ;) After a bit more experimentation, I realise that the interesting case is where the macro expands within another class. So, it should be sufficient to just instantiate `ParentClass` in the main body of the tests to get the test to a) run, and b) test the thing that's been fixed!

core/shared/src/main/scala/magnolia.scala
shadaj: When none of the parameters have default values, we can directly use the `constructorParameters` instead of having to get the `apply` method in the companion, which throws an exception when the case class is inside another class.

Do you know if it's possible to get the default parameter values without using the companion object?

propensive: Ah, I see. I don't think it's easy. There might be a better way than the one I'm using, though... StackOverflow may have the answer... ;)

joroKr21:
@shadaj The case classes in your tests are not inner classes, but local classes (their scope is constrained in a block passed to `test`). The bug is that `genericType.companion` returns `NoType` when `genericType` is a local class. If you want to avoid reinventing the wheel take look at [`patchedCompanionSymbolOf`](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/generic.scala#L698) in shapeless.

PR #42, opened 2017-11-20
Add TypeclassCompanion helper trait

leandrob13:
Adds the `TypeclassCompanion` helper trait that defines the properties needed for the derivation macro to work

```scala

trait TypeclassCompanion[C[_]] {

  /** Required type constructor for new instances of the typeclass */
  type Typeclass[T] = C[T]

  /** The typeclass constructor */
  def apply[T](implicit c: C[T]): C[T] = c

  /** Defines how a typeclass should be constructed for a given case class */
  def combine[T](ctx: CaseClass[C, T]): C[T]

  /** Defines how to choose which subtype of a sealed trait to use for the
    * construction of the typeclass
    * */
  def dispatch[T](ctx: SealedTrait[C, T]): C[T]

  /** binds the Magnolia macro to the derivation object */
  implicit def gen[T]: C[T] = macro Magnolia.gen[T]

}
```

How it works is shown in the updated examples.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target
 .jvm
 .js
 .native
+.idea
--- a/core/shared/src/main/scala/TypeclassCompanion.scala
+++ b/core/shared/src/main/scala/TypeclassCompanion.scala
@@ -1,0 +1,25 @@
+package magnolia
+
+import scala.language.experimental.macros
+import scala.language.higherKinds
+
+trait TypeclassCompanion[C[_]] {
+
+  /** Required type constructor for new instances of the typeclass */
+  type Typeclass[T] = C[T]
+
+  /** The typeclass constructor */
+  def apply[T](implicit c: C[T]): C[T] = c
+
+  /** Defines how a typeclass should be constructed for a given case class */
+  def combine[T](ctx: CaseClass[C, T]): C[T]
+
+  /** Defines how to choose which subtype of a sealed trait to use for the
+    * construction of the typeclass
+    * */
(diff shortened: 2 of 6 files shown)


propensive:
Hi @leandrob13, thank you for this!

I'm hesitant to merge it, though, as I'm not sure if it's too constraining: with the current no-interface approach, the macro can call the methods `combine` and `dispatch` regardless of whether they take additional implicit parameters, or how many type parameters they have. I don't yet have any use cases for this, but I'd like to be certain they don't exist before making this interface the "standard" way to create a derivation object.

I like that it would allow me to remove the logic in the macro which checks for the existence of `combine` and `dispatch`. So I'll keep the PR open for now.

propensive:
I've also noted that the trait could be used optionally, as guidance for the implementation, where the macro wouldn't pay any attention to it. I'm not currently sure if this would help a majority of users, or would introduce some confusion if users realize that the `... extends TypeclassCompanion[Foo]` is actually unnecessary. On balance, I think it's beneficial, but I'll think about it some more.

leandrob13:
@propensive no problem and totally understandable from that perspective. I just found myself using this several times in two different parser typeclass derivations that I thought it might be useful to have it in the lib not as a mandatory solution but as a guide for the lazy (like me) but you are right about giving the impression of limiting the parameters to only one.

propensive:
In issue #46 (almost ready), I generalize the interface for this quite a bit, by adding four more (optional) methods. I think that will make any attempt to "traitify" the "interface" near enough impossible, so I'm going to (reluctantly) close this PR, but I appreciate the work. Thank you!

PR #48, opened 2017-11-25
Fixed derivation error when companion has alternative apply methods

### LINKED ISSUE #47: Fails to derive instance for case classes with additional contructor in companion object
If the case class has an extra overloaded `apply` constructor in the companion object, it fails to derive an instance for that class. This example will fail:

```scala
case class Test(params: List[Params])

case class Params(a: String, b: String)

object Test {

  def apply(a: String, b: String): Test = Test(List(Params(a, b)))
}
```

leandrob13:
Fixes #47

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -279,15 +279,16 @@ object Magnolia {
 
         val defaults = if (!isValueClass) {
           val caseClassCompanion = genericType.companion
+
           // If a companion object is defined with alternative apply methods
           // it is needed get all the alternatives
           val constructorMethod =
-          caseClassCompanion.decl(TermName("apply")).alternatives.map(_.asMethod)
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/magnolia.scala
```diff
@@ -279,9 +279,16 @@ object Magnolia {
...
+          // If a companion object is defined with alternative apply methods
+          // it is needed get all the alternatives
+          val constructorMethods =
+            caseClassCompanion.decl(TermName("apply")).alternatives.map(_.asMethod)
+
+          // The last apply method in the alternatives is the one that belongs
+          // to the case class, not the user defined companion object
           val indexedConstructorParams =
-            constructorMethod.paramLists.head.map(_.asTerm).zipWithIndex
+            constructorMethods.last.paramLists.head.map(_.asTerm).zipWithIndex
```
propensive: Glad you commented there, or I'd have asked how you know the `last` one is the one we want! ;)

Issue #46, opened 2017-11-22
Support user-defined lifted types


propensive:
This feature has several advantages over v0.6.1:
1. Magnolia would no longer need to be a runtime dependency of any libraries which use it,
2. implicit evidence (of some type other than the typeclass being derived) could be requested for each parameter,
3. runtime and compile-time performance could be improved by reducing garbage generation

The macro currently instantiates `CaseClass`, `SealedTrait`, `Subtype` and `Param` instances which it passes to the user-defined method, `combine`. Magnolia provides fully-functional versions of these types. That is, including all the features which may be utilized in any given derivation, regardless of whether they are actually used or not.

However, there is scope for alternative implementations of these types to be used, and it should be possible for users to abstract over the instantiation of these types in general ways, for example by resolve additional typeclass instances on the type of the parameter (which is not currently possible from within the `combine` method definition, as the parameter type is existential by that point).

I propose to give users the option of providing their own implementations of these types, to be instantiated by new methods provided on the derivation object, namely,
- `param`,
- `caseClass`,
- `sealedTrait`, and
- `subtype`

As with `combine` and `dispatch`, the signatures of these methods would not have to adhere to a fixed interface (in the object-oriented sense), and could take additional type parameters or implicit parameters, but would be invoked by the macro which would inspect their parameters' names, and supply values only to those parameters in the signature. Any parameters (e.g. support for `default` values) not listed as the parameter would not be generated.

The first advantage would be that a number of ephemeral objects which exist only for covariant or only for contravariant typeclasses would not need to be instantiated for all cases, even when they are not used in the derivation. This should be beneficial for both compile time and runtime performance.

The second advantage would be that user-defined classes could be used in preference to Magnolia's default implementations, thereby obviating the need for Magnolia to be a runtime dependency. Given the library is likely to be useful for a number of other projects, the risk of a diamond dependency problem could be eliminated, while for application (i.e. non-library) developers, the tersity of the status quo would apply.

The existing API would remain unchanged if users did not provide any of the custom methods.

SystemFw:
Nice to see this being worked on :)

PR #53, opened 2017-11-29
Add support for repeated parameters

### LINKED ISSUE #24: Case classes with repeated parameters do not work
This should be tested, and the best solution is to probably treat the repeated parameters of type `T` as a single parameter of type `Seq[T]`.

joroKr21:
* A parameter of type `A*` is encoded as a `Seq[A]`
* Added a `Param.repeated` flag to `interface`, but I'm not sure if this is useful

Resolves #24


core/shared/src/main/scala/interface.scala
```diff
@@ -46,6 +46,17 @@ trait Param[Typeclass[_], Type] {
...
+  /** flag indicating a repeated (aka. vararg) parameter
+    *
+    * For example, for a case class,
+    * <pre>
+    * case class Account(id: String, emails: String*)
+    * </pre>
+    * the [[Param]] instance corresponding to the `emails` parameter would be `repeated` and have a
+    * [[PType]] equal to the type `Seq[String]`. Note that only the last parameter of a case class
+    * can be repeated. */
+  def repeated: Boolean
```
propensive: I'm happy for this additional parameter to be here. It's an extra parameter to the interface, but in #46 I'm going to offer a way to make that optional (which might result in a tiny runtime and compile-time speedup).

core/shared/src/main/scala/magnolia.scala
```diff
@@ -204,9 +206,9 @@ object Magnolia {
       val assignedName: TermName = TermName(c.freshName(s"${genericTypeName}Typeclass"))
       val typeSymbol = genericType.typeSymbol
       val classType = if (typeSymbol.isClass) Some(typeSymbol.asClass) else None
-      val isCaseClass = classType.map(_.isCaseClass).getOrElse(false)
-      val isCaseObject = classType.map(_.isModuleClass).getOrElse(false)
-      val isSealedTrait = classType.map(_.isSealed).getOrElse(false)
+      val isCaseClass = classType.exists(_.isCaseClass)
+      val isCaseObject = classType.exists(_.isModuleClass)
+      val isSealedTrait = classType.exists(_.isSealed)
```
propensive: Heh, no idea why this wasn't done before! ;)

tests/src/main/scala/tests.scala
```diff
@@ -176,5 +176,13 @@ object Tests extends TestApp {
     test("serialize a value class") {
       Show.gen[Length].show(new Length(100))
     }.assert(_ == "100")
+
+    test("show a case class with repeated param") {
```
propensive: Could you include a second test which handles a repeated parameter which is also a Magnolia-derived type, please? Not likely to be a problem, but an additional test would be useful.

joroKr21: I'm sorry I don't follow. What do you mean by Magnolia-derived type?

propensive: I mean a repeated type which needs Magnolia to derive the typeclass, i.e. not a `String` (`Seq[String]`), but a `Foo` (`Seq[Foo]`) for some case-class type, `Foo`.

joroKr21: Oh I see, will do :+1:

joroKr21:
done

PR #54, opened 2017-11-30
 Existentially abstract unbound subtype parameters 

### LINKED ISSUE #51: Unhelpful error for subtype with more type parameters
Maybe an edge case, but not so hard to deal with. And since one of the cool features of Magnolia is the error reporting here goes:
```scala
sealed trait Politician[A]
case class Accountable[A](always: A) extends Politician[A]
case class Corrupt[A, B](appearance: A, agenda: B) extends Politician[A]
Show.gen[Politician[String]] // key not found: B
```
I see two options:
1. Existential quantification - not very useful but perfectly sound
2. Throw a more descriptive error

joroKr21:
That happens when the subtype of a sealed trait has more type parameters than its parent. When those extra type parameters are covariant they are replaced by their upper bounds, otherwise they are existentially quantified.

Fixes #51

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1,3 +1,3 @@
 Jon Pretty <jon.pretty@propensive.com> [@propensive](https://twitter.com/propensive/)
 Loïc Descotte
-
+Georgi Krastev <jorokr21@gmail.com> [@joro_kr](https://twitter.com/joro_kr/)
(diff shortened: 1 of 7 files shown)


propensive:
Thank you for this! I hadn't expected a solution so fast (with so little work required from myself)! :)

I'll try to review and merge later today, but if all the old tests still pass, it's looking good.

PR #67, opened 2018-01-04
Change `typeName: String` to `typeName: magnolia.TypeName` [BREAKING!]

sirthias:
This proposes a breaking change introducing a more fine-grained structure to the type name information that magnolia provides.

Currently users only receive a `typeName: String` information containing the full class name.
Selecting only the owner part or only the short name requires brittle and inefficient manual string manipulation.

This proposal changes `typeName` to

```scala
case class TypeName(ownerName: String, short: String) {
  def full: String = s"$ownerName.$short"
}
```

As @propensive already hinted to [here](https://github.com/propensive/magnolia/pull/63#discussion_r158933243) the next release might be breaking the API anyway, which might be a good opportunity to sneak this change in, if deemed valuable.


sirthias:
Thank you, @propensive!
Personally, I am not worried about this one additional allocation *per type class instance* as long as the potentially hot code run *per type class invocation* is efficient.

However, I just realized that I forgot to upgrade the `Subtype#label` member to the new `TypeName`. I'll put in another PR with that change as well.

propensive:
Ok, thanks!

As an aside, I should really write some benchmarks...

PR #68, opened 2018-01-04
Rework compile time stack

### LINKED ISSUE #44: Derivation via a chained implicit does not report errors correctly
There are no test cases for this, but derivations which re-enter the Magnolia macro through implicit search via another implicit (e.g. `def showList[T: Show]: Show[List[T]] = ...`) do not appear to report this in the "stack trace" that's displayed when derivation fails.

### LINKED ISSUE #60: Derivation breaks when type classes depend on each other
Magnolia assumes the type class being derived is always the same as the root.
Say we change `Show.seq` (for demonstration's sake):
```scala
// Now we also require Default[A] - a different Magnolia-derived type class
implicit def seq[A](implicit A: Default[A], default: Default[A]): Show[String, Seq[A]] =
  // ... doesn't matter
```
Then we add another type of branch to `Tree`:
```scala
case class Many[+B](children: Tree[B]*) extends Tree[B]
```
Whoops, cache invalidation is hard:
```scala
Show.gen[Tree[String]]
// sbt:magnolia>
[error] .../magnolia/tests/src/main/scala/tests.scala:67:20: type mismatch;
[error]  found   : magnolia.examples.Show.Typeclass[magnolia.tests.Tree[String]]
[error]     (which expands to)  magnolia.examples.Show[String,magnolia.tests.Tree[String]]
[error]  required: magnolia.examples.Default[magnolia.tests.Tree[String]]
[error]   examples.Show.gen[Tree[String]]
```
Granted, the example above is contrived but situations like this can easily arise when mixing different libraries that use Magnolia for deriving type classes.

The fix would be something along the lines of using `Typeclass[T]` instead of `T` as the key everywhere.

### LINKED ISSUE #66: Derivation fails with a StackOverflowError at runtime
This fails with a StackOverflowError at runtime
```scala
  case class Recursive( r: Seq[Recursive] )
  println( magnolia.examples.Show.gen[Recursive].show( Recursive( List( Recursive( Nil) ) ) ) )
```

And surprisingly this does work, if a sealed trait is added on top of the case class.
```scala
  sealed trait Recursive2
  case class Recursive1( r: Seq[Recursive2] ) extends Recursive2
  println( magnolia.examples.Show.gen[Recursive2].show( Recursive1( List( Recursive1( Nil) ) ) ) )
```
And capturing the show instance as an implicit val produces a NullPointerException
```scala
  case class Recursive( r: Seq[Recursive] )
  implicit val showRecursive : Show[String,Recursive] = Show.gen[Recursive]
  println( showRecursive.show( Recursive( List( Recursive( Nil ) ) ) ) )
```

joroKr21:
There are a lot changes. Still need to rebase and resolve conflicts but should be possible to review like this.

Some design choices:
* Embrace a mutable stack (having a global case class `var allTheState` doesn't help, it just makes it cumbersome to modify without lenses).
* Add the typeclass constructor to stack frames and error messages (see #60)
* Also includes some clean-up (IMO) around `Option`s

fixes #44, fixes #60 and fixes #66

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -1,11 +1,9 @@
 package magnolia
 
-import scala.util.control.NonFatal
-import scala.reflect._
-import macros._
-import scala.collection.immutable.ListMap
-import language.existentials
-import language.higherKinds
+import scala.collection.mutable
+import scala.language.existentials
+import scala.language.higherKinds
+import scala.reflect.macros._
 
 /** the object which defines the Magnolia macro */
 object Magnolia {
@@ -65,7 +63,7 @@ object Magnolia {
     *  will suffice, however the qualifications regarding additional type parameters and implicit
     *  parameters apply equally to `dispatch` as to `combine`.
     *  */
-  def gen[T: c.WeakTypeTag](c: whitebox.Context): c.Tree = {
+  def gen[T: c.WeakTypeTag](c: whitebox.Context): c.Tree = Stack.withContext(c) { stack =>
     import c.universe._
     import internal._
 
@@ -73,13 +71,15 @@ object Magnolia {
       .find(_.tree.tpe <:< typeOf[debug])
       .flatMap(_.tree.children.tail.collectFirst { case Literal(Constant(s: String)) => s })
 
-    val magnoliaPkg = q"_root_.magnolia"
-    val scalaPkg = q"_root_.scala"
+    val magnoliaPkg = c.mirror.staticPackage("magnolia")
+    val scalaPkg = c.mirror.staticPackage("scala")
 
     val repeatedParamClass = definitions.RepeatedParamClass
     val scalaSeqType = typeOf[Seq[_]].typeConstructor
 
     val prefixType = c.prefix.tree.tpe
+    val prefixObject = prefixType.typeSymbol
+    val prefixName = prefixObject.name.decodedName
 
     val typeDefs = prefixType.baseClasses.flatMap { cls =>
       cls.asType.toType.decls.filter(_.isType).find(_.name.toString == "Typeclass").map { tpe =>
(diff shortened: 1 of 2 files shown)


propensive:
Thank you for this! I'm liking the look of it so far. Hopefully the rebase won't be too difficult...

joroKr21:
Rebased, but still need to add a test for #60

propensive:
Thanks. Looks like you'll need another small rebase now, after Mathias's changes were merged. Let me know when you're ready to go, and I'll quickly have a go at my #46 PR again, with quite a few more tests since last checking.

joroKr21:
@propensive I ran the "benchmarks", here is the result. There is a noticeable slowdown, but I guess caching can be added again in a separate PR.

* with caching:
```
00:07.290963   src/adt.scala
00:14.905706   src/magnolia/show_50.scala
00:07.488627   src/magnolia/show_50.scala
00:05.804537   src/magnolia/show_50.scala
00:04.834441   src/magnolia/show_50.scala
00:04.626380   src/magnolia/show_50.scala
00:04.271303   src/magnolia/show_50.scala
00:04.004326   src/magnolia/show_50.scala
00:03.537197   src/magnolia/show_50.scala
00:03.635504   src/magnolia/show_50.scala
00:03.851901   src/magnolia/show_50.scala
```
* with the reworked stack:
```
00:06.950972   src/adt.scala
00:17.466868   src/magnolia/show_50.scala
00:08.859871   src/magnolia/show_50.scala
00:06.330354   src/magnolia/show_50.scala
00:05.912984   src/magnolia/show_50.scala
00:05.277233   src/magnolia/show_50.scala
00:05.011331   src/magnolia/show_50.scala
00:04.728933   src/magnolia/show_50.scala
00:04.456627   src/magnolia/show_50.scala
00:04.430862   src/magnolia/show_50.scala
00:04.753720   src/magnolia/show_50.scala
```

joroKr21:
I added caching again. Not sure if it helps much. Also added a test for dependencies between derived type classes. This should now be ready.

PR #69, opened 2018-01-04
Fix regression introduced by move to `patchedCompanionRef`

sirthias:
[This commit](https://github.com/propensive/magnolia/commit/c91b7cac849237062eea8a75134783e826dc21a0) introduced a regression that broke things like `Show.gen[List[Int]]`.

This patch selectively rolls back the changes from that commit, which green-lights the added test.

I have to say that I have absolutely no insight whatsoever into the intricacies of the name resolution here, so a proper review by @propensive and @joroKr21 is probably of the essence.

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -225,11 +225,10 @@ object Magnolia {
       }
 
       val result = if (isCaseObject) {
-        val obj = GlobalUtil.patchedCompanionRef(c)(genericType)
         val impl = q"""
           $typeNameDef
           ${c.prefix}.combine($magnoliaPkg.Magnolia.caseClass[$typeConstructor, $genericType](
-            $typeName, true, false, new $scalaPkg.Array(0), _ => $obj)
+            $typeName, true, false, new $scalaPkg.Array(0), _ => ${genericType.typeSymbol.asClass.module})
           )
         """
         Some(Typeclass(genericType, impl))
--- a/tests/src/main/scala/tests.scala
+++ b/tests/src/main/scala/tests.scala
@@ -318,6 +318,10 @@ object Tests extends TestApp {
     test("show a Portfolio of Companies") {
       Show.gen[Portfolio].show(Portfolio(Company("Alice Inc"), Company("Bob & Co")))
     }.assert(_ == "Portfolio(companies=[Company(name=Alice Inc),Company(name=Bob & Co)])")
+
+    test("show a List[Int]") {
+      Show.gen[List[Int]].show(List(1, 2, 3))
+    }.assert(_ == "::(head=1,tl$access$1=::(head=2,tl$access$1=::(head=3,tl$access$1=Nil())))")
     
     test("sealed trait typeName should be complete and unchanged") {
       TypeNameInfo.gen[Color].name


joroKr21:
Oh, I see what's happening. `List` is actually a type alias defined in package object `scala`. It has no companion. A better solution might be to call `dealias` either in `patchedCompanionRef` itself or at all call sites of it.

propensive:
I think @joroKr21 will need to check this, and give me the :+1:. I don't know enough about the original problem to fix it.

joroKr21:
About the underlying problem: it's a bug related to local classes (classes nested in a method, val, any block really) - for some reason they are not treated as companions. For example:
```scala
def foo(): Unit = {
  class Bar
  object Bar // not a companion
}
```
I'm not sure if this is only during macro expansion or goes on like this to the backend. So the patched lookup is just a best effort to find the companion. Personally, I use local classes only for testing. I would even be in favour of deleting `pathedCompanionRef` altogether and issue an error that says local classes are not supported...

sirthias:
Thank you for your comments, @joroKr21!
Unfortunately the `.dealias` alone doesn't fix the problem, i.e. the macros generates the same (broken) code as without the `.dealias`.

This is the compiler error:

```
[error] .../magnolia/tests/src/main/scala/tests.scala:323:15: object package is not a member of package scala.collection.immutable
[error]       Show.gen[List[Int]].show(List(1, 2, 3))
[error]               ^
```

And this is the offending code generated by the macro:

```scala
magnolia.examples.Show.combine(
  _root_.magnolia.Magnolia.caseClass[magnolia.examples.Show.Typeclass, scala.collection.immutable.Nil.type](
    "scala.collection.immutable.Nil",
    true,
    false,
    new _root_.scala.Array(0), ((x$128) => scala.collection.immutable.`package`.<none>)
  )
)
```

propensive:
Perhaps it would be better to just not support local classes. I like the idea of deleting code, and this would remove a significant portion of the total, not least because it's depending on compiler internals.

I wouldn't be averse to that as a solution for *magnolia*, resting on the hope that hack could be integrated into the compiler, with Olaf's work on macros. Maybe I'm optimistic.

sirthias:
Yes, I think dropping support for local classes shouldn't be too much of a problem for most users.

However, simple "forwarders", like

```scala
type List[+A] = scala.collection.immutable.List[A]
val List = scala.collection.immutable.List
```

on `Predef` should just work, IMHO.
If it's too hard/brittle to do generically we might even want to special case all the `Predef` stuff.

joroKr21:
Whoops, there is another (unrelated) problem here. When the type is a `case object` we don't want to get the `companion` (there is none), we want the `module`.

core/shared/src/main/scala/magnolia.scala
```diff
@@ -225,7 +225,14 @@ object Magnolia {
       }
 
       val result = if (isCaseObject) {
-        val obj = GlobalUtil.patchedCompanionRef(c)(genericType)
+        val obj = {
```
sirthias: Maybe a simple `val obj = genericType.typeSymbol.name.toTermName` would suffice here?

joroKr21: I think `val obj = genericType.typeSymbol.asClass.module` is most precise

sirthias: Ok, great.
I'll go for that then here.
Thank you!

sirthias:
Ok, updated to the latest suggestion by @joroKr21.

propensive:
Two PRs merged this year... you're on a roll, @sirthias! Thanks for the help along the way, @joroKr21. :)

PR #70, opened 2018-01-05
Add support for `magnolia.debug` annotation

sirthias:
This addition has proven helpful on my side to be able to quickly diagnose problems with magnolia's macro expansion.

I think it'd be valuable to ship with the next release in order to have a mechanism for receiving meaningful debugging information from users in the field.


sirthias:
Great!
Thank you for being so open towards random contributions! :)

PR #85, opened 2018-03-29
Add option flag to parameters

jodersky:
In certain situations, parameters of type Option have a special meaning, usually symbolizing the possible absence of a value in a serialized form. To allow deserialization in these situations, it is required to introspect a parameter's type and in case it is of type Option, allow for the absence of that parameter in the actual deserialized value.
One such example is JSON typeclass derivation. Typically, an option parameter of value None may be omitted in the serialization of objects and will be equivalent to json "null" during deserialization. Unfortunately, it is currently impossible to query the type of a parameter in magnolia, hence making it [impossible to deserialize JSON objects that do not set fields for every Option parameter](https://github.com/drivergroup/spray-json-derivation/issues/5) of their case case class.

This patch adds a boolean field "option" to parameters, which is set if the parameter is of type option. See [here](https://github.com/drivergroup/spray-json-derivation/blob/option/src/main/scala/DerivedFormats.scala#L28) for an example use-case (or [this diff](https://github.com/drivergroup/spray-json-derivation/compare/option) for more context).

It may also be helpful to expose parameters' types instead of the the specific option query proposed, however I'm not sure how a `Type` from the macro realm can be passed along to userland. I'd be open to suggestions to work on that too.


jodersky:
what do you think of this proposal in general? Anything that needs clarification or refactoring?

propensive:
Sorry it's taken me so long to notice this PR. For some reason I've not been getting notifications from Magnolia... (and I haven't been proactively checking!)

I think it's essential we support these sorts of derivations. Though I'm wondering whether there is a way to do this which doesn't special-case `Option`; and I completely get your point about generalizing it to other types (but not having a good way to represent these). There would be a tiny additional runtime performance cost for each `isOption` field for every `Param` object.

So I'm going to think about this some more. Ideally, the special casing for `Option`s would be pushed out into a special chained typeclasses for `Option`s, along the lines of,
```
implicit def option[T: Typeclass]: Typeclass[Option[T]] = ...
```
Not sure if that would apply here...

jodersky:
I'm not sure I quite understand what you mean by chained typeclasses for `Option`s. In the scenario described, we'd like to know if a field of a case class is an option so that we can modify the serialization of the case class itself (not the field). E.g.
```scala
case class A(x: Option[Int])

// without option field serialization, this would fail since the fiel "x" is not defined:
"{}".parseJson.convertTo[A]

// with option field serialization, this works
"{}".parseJson.convertTo[A] == A(None)
```

fommil:
I don't see why `Option` should be treated as special... it feels to me like the annotations are more appropriate for customisations.

jodersky:
My reasoning was that some data formats (such as JSON)  allow sending null values by omitting fields entirely. To be able to deserialize them we need to inspect those omittable values in the parent object, before dereferencing them.
Such omittable fields could very well be defined with annotations, however I think using option types to represent such fields is very convenient and also a fairly common pattern in Scala.

fommil:
If you need to special case a typeclass then pattern match on it :shrug:

fommil:
also, if you want to special case a type why not just compare against the `TypeName`?

I may do something similar in an example json library to distinguish between "null and missing" during decoding.

propensive:
Happy to reopen if necessary, but I think the solution @fommil describes should work, so I'll close this.

PR #93, opened 2018-05-11
use toList instead of to[List]

xuwei-k:
prepare Scala 2.13.0-M4

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -333,7 +333,7 @@ object Magnolia {
         }})}))
           }""")
       } else if (isSealedTrait) {
-        val genericSubtypes = classType.get.knownDirectSubclasses.to[List]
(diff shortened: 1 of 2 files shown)


xuwei-k:
@propensive

propensive:
Thanks for the nudge, @xuwei-k - I didn't see the original notification for some reason...

I haven't tried playing around with Scala 2.13 collections yet, but am I right in understanding that `.to[List]` is deprecated (or removed?). I always preferred it for its uniformity across collections. Is the new "uniform" version `.to(List)`? I would prefer that, but this is bikeshedding noise so I won't hold up merging!

SethTisue:
yes, the new uniform method in 2.13 is `.to(List)`, but in order for usages of `.to(List)` to cross-build on 2.11/2.12/2.13, you would need to add a dependency on scala-collection-compat. it's better to avoid the extra dependency if you can, hence just using `.toList` in this particular case is appropriate IMO

PR #98, opened 2018-06-05
Calculate all leaf nodes for knownDirectSubclasses

### LINKED ISSUE #88: [Circe] Add a way to skip dispatch over intermediate abstract classes/traits
Reproduction: https://github.com/vpavkin/magnolia/blob/e278256e713ddf3b1aa3365898e930eff168274b/tests/src/main/scala/IntermediateTraitsTest.scala

Shapeless generic skips intermediate traits/abstract classes and forms a coproduct of only leaf types.
This makes perfect sense for many scenarios, JSON included.

Magnolia, on the other hand, dispatches through all intermediate types.
For encoding it's not that bad, but for decoding it's a showstopper.

Given
```scala
  sealed trait T
  case class A(a: Int) extends T
  case class B(b: String) extends T
  sealed trait C extends T
  case class C1(c1: Int) extends C
  case class C2(c2: String) extends C
```

Encoding an instance of `C1` as `T` will produce this (which is not nice, but manageable probably):
```
{
   "C":{
      "C1":{
         "c1":2
      }
   }
}
```

But problem is that to properly decode an instance of `C1` we have to scope it as `C`, which is very unfriendly to JSON API users. And the more simple and intuitive form won't decode:
```
{
   "C1":{
      "c1":2
   }
}
```

propensive:
Might fix #88.

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -99,6 +99,13 @@ object Magnolia {
 
     def error(msg: String) = c.abort(c.enclosingPosition, msg)
 
+    def knownSubclasses(sym: ClassSymbol): List[Symbol] = {
+      val children = sym.knownDirectSubclasses.toList
+      val (abstractTypes, concreteTypes) = children.partition(_.isAbstract)
+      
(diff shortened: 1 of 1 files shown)


vpavkin:
Confirm, my test case for #88 passed with this :+1:

PR #108, opened 2018-06-17
Use strict vals by default and fall back to lazy vals

joroKr21:
when subtrees are Deferred or reference enclosing vals.


propensive:
Looks great, thank you! Sorry it took me so long to merge...

PR #116, opened 2018-07-24
checkMethod: termName is not used in the error message

iRevive:
Method `checkMethod` is not using `termName` in the error message.


propensive:
Thank you!

PR #120, opened 2018-08-03
econstruct, like scalaz.MonadError.emap

### LINKED ISSUE #118: CaseClass.constructEither
The current implementation of `CaseClass.construct` means that typeclasses that stop at `Functor` can be implemented (e.g. `Default`), but typeclasses with a `MonadError` (e.g. a JSON decoder) cannot.

The workaround is to use locally scoped exceptions, e.g.:

```scala
  private case class Fail(msg: String) extends Exception with NoStackTrace

  def combine[A](ctx: CaseClass[JsDecoder, A]): JsDecoder[A] = {
    case obj @ JsObject(_) =>
      try {
        val success = ctx.construct { p =>
          val value = obj.get(p.label).getOrElse(JsNull)
          p.typeclass.fromJson(value) match {
            case \/-(got)  => got
            case -\/(fail) => throw new Fail(fail) // scalafix:ok
          }
        }
        \/-(success)
      } catch {
        case Fail(msg) => -\/(msg)
      }
    case other => JsDecoder.fail("JsObject", other)
  }
```

it would be better if you allowed a traversal, even just for `Either`, allowing

```scala
  private case class Fail(msg: String) extends Exception with NoStackTrace

  def combine[A](ctx: CaseClass[JsDecoder, A]): JsDecoder[A] = {
    case obj @ JsObject(_) =>
        ctx.constructEither { p =>
          val value = obj.get(p.label).getOrElse(JsNull)
          p.typeclass.fromJson(value)
        }
    case other => JsDecoder.fail("JsObject", other)
  }
```

If you produced the boilerplate that called `.flatMap` then users could even use their own data types, not just `Either`.

fommil:
close #118

### DIFF
--- a/core/shared/src/main/scala/interface.scala
+++ b/core/shared/src/main/scala/interface.scala
@@ -160,16 +160,19 @@ abstract class CaseClass[Typeclass[_], Type] private[magnolia] (
    *
    * @see construct
    */
-  final def econstruct[E <: AnyRef, Return](makeParam: Param[Typeclass, Type] => Either[E, Return]): Either[E, Type] = {
+  final def constructEither[E <: AnyRef, Return](makeParam: Param[Typeclass, Type] => Either[E, Return]): Either[E, Type] = {
     // poor man's scalaz.Traverse
     try {
-      val bits = parameters.map { p =>
-        makeParam(p) match {
-          case Left(e) => throw EarlyExit(e)
-          case Right(a) => a
-        }
-      }
-      Right(rawConstruct(bits))
+      Right(
+        rawConstruct(
+          parameters.map { p =>
+            makeParam(p) match {
+              case Left(e) => throw EarlyExit(e)
+              case Right(a) => a
+            }
+          }
+        )
+      )
     } catch {
       case EarlyExit(err) => Left(err.asInstanceOf[E])
     }
--- a/core/shared/src/main/scala/interface.scala
+++ b/core/shared/src/main/scala/interface.scala
@@ -155,6 +155,26 @@ abstract class CaseClass[Typeclass[_], Type] private[magnolia] (
   final def construct[Return](makeParam: Param[Typeclass, Type] => Return): Type =
     rawConstruct(parameters map makeParam)
 
+  /**
+   * Like construct but allows each parameter to fail with an error.
+   *
+   * @see construct
+   */
+  final def econstruct[E <: AnyRef, Return](makeParam: Param[Typeclass, Type] => Either[E, Return]): Either[E, Type] = {
+    // poor man's scalaz.Traverse
+    try {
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/interface.scala
```diff
@@ -155,6 +155,26 @@ abstract class CaseClass[Typeclass[_], Type] private[magnolia] (
...
+   * Like construct but allows each parameter to fail with an error.
+   *
+   * @see construct
+   */
+  final def econstruct[E <: AnyRef, Return](makeParam: Param[Typeclass, Type] => Either[E, Return]): Either[E, Type] = {
+    // poor man's scalaz.Traverse
+    try {
+      val bits = parameters.map { p =>
+        makeParam(p) match {
+          case Left(e) => throw EarlyExit(e)
```
propensive: `return Left(err.asInstanceOf[E])`, and remove the `try`/`catch`?

core/shared/src/main/scala/interface.scala
```diff
@@ -155,6 +155,26 @@ abstract class CaseClass[Typeclass[_], Type] private[magnolia] (
...
 
+  /**
+   * Like construct but allows each parameter to fail with an error.
+   *
+   * @see construct
+   */
+  final def econstruct[E <: AnyRef, Return](makeParam: Param[Typeclass, Type] => Either[E, Return]): Either[E, Type] = {
+    // poor man's scalaz.Traverse
+    try {
+      val bits = parameters.map { p =>
```
propensive: `bits` is misleading, I think. How about writing it as a point-free expression as a way to avoid coming up with a unique name? ;)

core/shared/src/main/scala/interface.scala
```diff
@@ -248,3 +268,5 @@ final case class TypeName(owner: String, short: String) {
   *                     whose full name contains the given [[String]]
   */
 final class debug(typeNamePart: String = "") extends scala.annotation.StaticAnnotation
+
+private[magnolia] final case class EarlyExit[E](e: E) extends Exception with util.control.NoStackTrace
```
propensive: I think that using `return` will avoid the stacktrace being generated too.

core/shared/src/main/scala/interface.scala
propensive: I also don't think the name `econstruct` is clear enough. The `e` could stand for "either" or "exception". Let's go with `eitherConstruct`, unless you can suggest anything better...

core/shared/src/main/scala/interface.scala
fommil: doesn't work, because the nearest method seems to be the map closure (I tried it)

fommil:
@propensive changes made

PR #121, opened 2018-08-03
don't assume fullauto derivation

### LINKED ISSUE #119: non-implicit macro is descending where it shouldn't
I have created a magnolia deriver for pureconfig https://github.com/pureconfig/pureconfig/issues/396#issuecomment-410233339

Unless I special case the value class (which I don't want to do!) then the following data structure will derive the `ServiceName` 

```scala
@deriving(ConfigReader)
final case class LoggingConfig(serviceName: ServiceName)

final case class ServiceName(value: String) extends AnyVal
```

Whereas I expect a compile time error reporting a missing `ConfigReader[ServiceName]`. The user can then choose to use `@deriving` or `@xderiving` on `ServiceName`, the former treating it as a product, the latter using the typeclass's definition of `xmap`. I don't want magnolia deciding to special case `AnyVal` for me as there is a choice to be made by the user and this leads to typeclass decoherence.

fommil:
close #119

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -134,6 +134,8 @@ object Magnolia {
 
     checkMethod("combine", "case classes", "CaseClass[Typeclass, _]")
 
+    val fullauto = c.macroApplication.symbol.isImplicit
+
     val expandDeferred = new Transformer {
       override def transform(tree: Tree) = tree match {
         case q"$magnolia.Deferred.apply[$_](${Literal(Constant(method: String))})"
@@ -167,7 +169,10 @@ object Magnolia {
         stack.recurse(frame, searchType) {
           Option(c.inferImplicitValue(searchType))
             .filterNot(_.isEmpty)
-            .orElse(directInferImplicit(genericType, typeConstructor))
+            .orElse(
+              if (fullauto) directInferImplicit(genericType, typeConstructor)
+              else None
+            )
             .getOrElse {
               val missingType = stack.top.fold(searchType)(_.searchType)
               val typeClassName = s"${missingType.typeSymbol.name.decodedName}.Typeclass"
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/magnolia.scala
```diff
@@ -134,6 +134,12 @@ object Magnolia {
 
     checkMethod("combine", "case classes", "CaseClass[Typeclass, _]")
 
+    val fullauto = c.prefix.tree.tpe.baseClasses
+      .map(_.asType.toType.decl(TermName("gen")))
+      .find(gen => gen != NoSymbol)
+      .getOrElse(error(s"magnolia: the method get must be defined on the derivation $prefixObject"))
```
propensive: `s/get/gen/`

core/shared/src/main/scala/magnolia.scala
```diff
@@ -134,6 +134,12 @@ object Magnolia {
 
     checkMethod("combine", "case classes", "CaseClass[Typeclass, _]")
 
+    val fullauto = c.prefix.tree.tpe.baseClasses
+      .map(_.asType.toType.decl(TermName("gen")))
```
propensive: So, right now, the method does not have to be called `gen`. Could we instead check whether the method being invoked (whatever it's called) is `implicit` or not?

propensive: I think `c.macroApplication` might help.

fommil: I asked about this in `scala/contributors` because it has stumped me. I'll try that out.

fommil: nice! That's exactly it.

joroKr21:
That's a bit of unintuitive change in behavior depending on the `implicit` flag. It should be well documented to avoid confusion. E.g. now I could still do "deep semiauto derivation" by making my macro implicit, however not importing it but calling it explicitly:

```scala
object FooDerivation {
  implicit def deriveFoo[T]: Foo[T] = macro ...
}

case class Bar(str: String)
case class Qux(bar: Bar)
FooDerivation.deriveFoo[Qux]
```

But how am I supposed to know that as a user?

fommil:
well it's true that I should document it. I'll do that in a follow up.

PR #123, opened 2018-08-03
semiauto derivation if gen is not implicit

fommil:
The motivation for this PR is that it allows one to derive an ADT but without it escaping into other people's types. Combined with the `@deriving` annotation it means we can write this

```scala
@deriving(Show, Equal, Arbitrary)
sealed abstract class JsValue
final case object JsNull                                    extends JsValue
final case class JsObject(fields: IList[(String, JsValue)]) extends JsValue
final case class JsArray(elements: IList[JsValue])          extends JsValue
final case class JsBoolean(value: Boolean)                  extends JsValue
final case class JsString(value: String)                    extends JsValue
final case class JsDouble(value: Double)                    extends JsValue
final case class JsInteger(value: Long)                     extends JsValue
```

instead of having to derive for all the products individually (and making their instances visible!)

```scala
@deriving(Show, Arbitrary)
final case object JsNull extends JsValue
@deriving(Show, Arbitrary)
final case class JsObject(fields: IList[(String, JsValue)]) extends JsValue
@deriving(Show, Arbitrary)
final case class JsArray(elements: IList[JsValue]) extends JsValue
@deriving(Show, Arbitrary)
final case class JsBoolean(value: Boolean) extends JsValue
@deriving(Show, Arbitrary)
final case class JsString(value: String) extends JsValue
@deriving(Show, Arbitrary)
final case class JsDouble(value: Double) extends JsValue
@deriving(Show, Arbitrary)
final case class JsInteger(value: Long) extends JsValue
```

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -136,6 +136,10 @@ object Magnolia {
 
     val fullauto = c.macroApplication.symbol.isImplicit
 
+    // halp! I really just want to ask "is t a subtype of T?"
+    def halfauto(s: Type): Boolean =
+      s.typeSymbol.isClass && s.typeSymbol.asClass.baseClasses.contains(weakTypeOf[T].typeSymbol)
+
     val expandDeferred = new Transformer {
       override def transform(tree: Tree) = tree match {
         case q"$magnolia.Deferred.apply[$_](${Literal(Constant(method: String))})"
@@ -170,7 +174,8 @@ object Magnolia {
           Option(c.inferImplicitValue(searchType))
             .filterNot(_.isEmpty)
             .orElse(
-              if (fullauto) directInferImplicit(genericType, typeConstructor)
+              if (fullauto || halfauto(genericType))
+                directInferImplicit(genericType, typeConstructor)
               else None
             )
             .getOrElse {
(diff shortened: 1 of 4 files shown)


fommil:
@joroKr21 something we could use to define the semantics would be the name of the `gen` method. `implicit` could imply full auto, then if it is called `semiauto` it would never recurse, otherwise do this "half auto":

```
- `implicit def gen` - full auto
- `def gen` - half auto
- `def semiauto` - semi auto only, require each product to be derived separately
```

propensive:
I'd prefer not to add significance to the name of the method. I'm not sure what would (or should) happen under a renaming import. And it would be surprising if the behaviour changed when a name changed. There is enough magic going on with "finding" the methods on the derivation object, anyway. ;)

core/shared/src/main/scala/magnolia.scala
```diff
@@ -136,6 +139,10 @@ object Magnolia {
 
     val fullauto = c.macroApplication.symbol.isImplicit
 
+    // halp! I really just want to ask "is t a subtype of T?"
+    def halfauto(s: Type): Boolean =
```
propensive: I don't think this works in the general case. Can't you use the `<:<` method?

fommil: the "halp" was definitely directed at you :smile: I'll try that...

fommil:
@propensive updated

PR #126, opened 2018-08-04
Subtype.annotations

### LINKED ISSUE #124: SubType.annotations
the only thing missing the `def annotations` is the Subtype and that means we can't have, for example, custom values instead of the `TypeName` for encoders and decoders.

fommil:
close #124 // @kevinwright quick review pls?


core/shared/src/main/scala/interface.scala
```diff
@@ -40,6 +40,10 @@ trait Subtype[Typeclass[_], Type] extends Serializable {
   /** partial function defined the subset of values of `Type` which have the type of this subtype */
   def cast: PartialFunction[Type, SType]
 
+  /** all of the annotations on the sub type */
+  final def annotations: Seq[Any] = annotationsArray
```
propensive: This is unrelated to your changes, but could we use `StaticAnnotation` instead of `Any`?

fommil: Yeah I'd prefer that

fommil:
@propensive done

fommil:
the only problem I can see with `StaticAnnotation` is that perhaps if there are non-static annotations the derivation may fail? // @kevinwright might know more

fommil:
@propensive ok, that was the problem... but perhaps using `Annotation` would be better? I don't think java annotations implement any of these interfaces.

PR #130, opened 2018-08-09
Construct monadic

propensive:


core/shared/src/main/scala/magnolia.scala
```diff
@@ -229,14 +230,23 @@ object Magnolia {
...
             new $scalaPkg.Array(0),
-            $scalaPkg.Array(..$classAnnotationTrees),
-            _ => ${genericType.typeSymbol.asClass.module}
-          ))
+            $scalaPkg.Array(..$classAnnotationTrees)
+          ) {
+            override def construct[Return](makeParam: _root_.magnolia.Param[$typeConstructor, $genericType] => Return): $genericType =
+              ${genericType.typeSymbol.asClass.module}
+
+            import _root_.scala.language.higherKinds
```
fommil: I suspect this is the thing causing the warning

PR #133, opened 2018-08-10
Made GenericShow generic so that is it not forced to be GenericShow[String]

anttiahola:
I was trying to understand the examples when I noticed that the `GenericShow` is actually not generic at all but forced to be `GenericShow[String]`. So I made a minor modification to it so that the generic part of the code is in `GenericShow` and the implementation for `GenericShow[String]` is in `Show`.

I verified this by making `ShowWeakHash`, based on the hash.scala -example, which is of type `GenericShow[Int]`.

I also changed the stringifying of the annotations in order to separate the case when the annotation is on the type versus when the annotation is on the parameter. So, the annotation tests were changed a bit also.

### DIFF
--- a/examples/shared/src/main/scala/show.scala
+++ b/examples/shared/src/main/scala/show.scala
@@ -27,40 +27,44 @@ trait GenericShow[Out] {
 
   /** the type constructor for new [[Show]] instances
     *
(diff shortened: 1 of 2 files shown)


joroKr21:
Is this PR still relevant?

joroKr21:
This PR is really outdated. Please reopen if you manage to update it.

PR #134, opened 2018-08-10
Make CallByNeed Serializable and Typename.full a val

jto:
Maybe this should be 2 PR but given how simple it is maybe not.

- Make CallByNeed `Serializable` is pretty much self explanatory
- `Make TypeName.full a val` is for perf reason.  It gets expensive if called repeatedly.


jto:
Not on it specifically but was running a microbenchmark in scio after upgrading to 0.10.0 and ~30% on the time was spent in this method.

PR #136, opened 2018-08-24
Ignore Java annotations, as we can't construct them (fixes #135)

### LINKED ISSUE #135: Derived typeclasses for types with Java annotations fail at runtime
For instance, annotating a case class with javax.annotation.Resource produces

```
[error] java.lang.InstantiationError: javax.annotation.Resource
[error] 	at magnolia.tests.Tests$$anon$72.<init>(tests.scala:443)
[error] 	at magnolia.tests.Tests$.$anonfun$tests$439(tests.scala:443)
```

This happens because the derived tree includes ``new javax.annotation.Resource()`` which won't work.

I can't find a better fix than to filter out Java annotations from the type annotations; a pull request to that effect will follow shortly.

espenhw:
As mentioned in #135, I think the only useful thing we can do with Java annotations is ignore them so they don't cause runtime failures.

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -100,6 +100,9 @@ object Magnolia {
     val prefixObject = prefixType.typeSymbol
     val prefixName = prefixObject.name.decodedName
 
+    val javaAnnotationType = typeOf[java.lang.annotation.Annotation]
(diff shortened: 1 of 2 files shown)


sirthias:
Is there anything holding this back?
#135 is actually causing issues in downstream projects, e.g. https://github.com/sirthias/borer/issues/13

kubukoz:
Superseded by #136 (thanks @espenhw!)

PR #158, opened 2019-03-30
Use a rarely used name for the monad type parameter in constructMonadic

adamw:
Using `F[_]` causes a warning:

```
type parameter F defined in method constructMonadic shadows type F defined in class (...). You may want to rename your type parameter, or possibly remove it.
```

each time magnolia is invoked in a context which is parametrised using `F[_]` itself.


jto:
Thanks Adam :)

adamw:
Np :) do you know if there are any plans for a release? I could also use the type parameter support ;)

joroKr21:
Sorry for the delayed response but a better solution might be to use `TypeName(c.freshName("F"))`

adamw:
@joroKr21 Nothing's lost! :) See https://github.com/propensive/magnolia/pull/159

PR #166, opened 2019-06-06
Support Java annotations by ignoring them!

### LINKED ISSUE #103: fields with annotations lead to "no arguments allowed for nullary constructor"
I am trying to use magnolia for the sphere-json and sphere-mongo libs.
We provide annotations on case class fields.
Ex:
```
  case class AnotherThing(
    @MongoKey("first_name") firstName: String
  )
```

By doing that, I see the following error:
```
no arguments allowed for nullary constructor MongoKey: ()io.sphere.mongo.generic.annotations.MongoKey
[error] Note that 'value' is not a parameter name of the invoked method.
[error]       val testFormat: MongoFormat[AnotherThing] = MongoFormatDerivation.gen
```

https://github.com/sphereio/sphere-scala-libs/blob/try_magnolia/mongo/src/test/scala/io/sphere/mongo/Serialization2Test.scala#L56

Any idea/hint how I could solve that? Or is it a limitation from magnolia?

kubukoz:
Java annotations are instantiated in a different way than Scala annotations (which are usually case classes, unlike Java annotations which are interfaces) - by overriding interface methods, instead of passing constructor parameters.

I tried doing that instead of passing the whole annotation call's tree, but apparently something in Scala makes it impossible to instantiate a Java annotation and actually grab an instance.

Should fix #103

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -211,7 +211,9 @@ object Magnolia {
             m.isPrivate
       }.getOrElse(false)
 
-      val classAnnotationTrees = typeSymbol.annotations.map(_.tree).filterNot(_.tpe.typeSymbol.isJavaAnnotation)
+      val isJavaAnnotation: Tree => Boolean = _.tpe.typeSymbol.isJavaAnnotation
+
+      val classAnnotationTrees = typeSymbol.annotations.map(_.tree).filterNot(isJavaAnnotation)
 
       val primitives = Set(typeOf[Double],
                            typeOf[Float],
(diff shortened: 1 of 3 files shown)


kubukoz:
Not really. IntelliJ magic...

kubukoz:
Having been informed that #136 exists, I decided to merge these changes here, so that this PR can supersede the other one without having @espenhw update their branch.

yanns:
Does this change mean that Java annotations are not accessible under `subType.annotations` for example?

kubukoz:
I think they *never were* accessible. They would just crash compilation at some point... but then again, this was some time ago and I'm not so sure. Are you having issues with that?

yanns:
We have a [library](https://github.com/sphereio/sphere-scala-libs/tree/master/json) using our own macros for derivation, and using Java annotations.
And I'm trying to switch to magnolia here.
Not having access to the Java annotations make the switch more difficult, but it's also an opportunity to get rid of those. So no blocker for me.
I just wanted to be sure before going the hard path of moving to scala `StaticAnnotation`.
Thx for the quick answer.

kubukoz:
If I remember correctly, the problem was that you can't instantiate a Java `@interface` from Scala code, so they couldn't be passed to the implementation of `dispatch` / `combine`. So yeah, it's more Scala's fault... so I guess `StaticAnnotation` is the way to go.

PR #179, opened 2019-09-29
Pop the stack on chained implicits and accumulate errors

### LINKED ISSUE #154: magnolia crashes scala on missing Option implicits
I had a missing implicit for `Option[Int]` for a serialiser typeclass I'm writing, 
instead of failing the implicit resolution for the top level data type that includes it, I got 2535 lines of scala compiler error, summarised by:

```
[error] (Compile / compileIncremental) java.lang.AssertionError: assertion failed:                                                                                            
[error]   new SerialVersionUID(value = 5066590221178148012L)                                                                                                                  
[error]      while compiling: <...>                    
[error]         during phase: superaccessors
[error]      library version: version 2.12.8
[error]     compiler version: version 2.12.8
```

Once the `Option[A]` implicit has been added this was fixed.

Looking at those many lines, it does seem that it tries to auto-magic it somehow?

```
TREE: magnolia.Magnolia.subtype[com.svalaskevicius.<...>.JsonSerialiser.Typeclass, Option[Int], None.type](magnolia.TypeName.apply("scala", "None"), 0, scala.Array.apply[Any](new SerialVersionUID(value = 5066590221178148012L))((ClassTag.Any: scala.reflect.ClassTag[Any])), magnolia.CallByNeed.apply[com.svalaskevicius.<...>.JsonSerialiser[None.type]]({
```

### LINKED ISSUE #178: Fails to derive typeclass instance for recursive class when referenced
Given recursive class RPerson - 
`case class RPerson(age: Int, name: String, child: Option[RPerson])`
derivation will work.

But it will fail as soon as we create another class which references the recursive class wrapping the reference with the same container as is used within the recursive class, e.g.
`case class GPerson(child: Option[RPerson])`

It doesn't matter if the container is an option/either/list or anything like that, but it has to be the same in both recursive class and the other one.

Here is a gist with the full example: https://gist.github.com/ghostbuster91/1e09a8d9fa9ca3d2ff99df0db732bdc4

joroKr21:
When we fail to find a chained implicit in `typeclassTree`
we still want to pop the stack, otherwise the logic is broken.
On the other hand we want to keep the failed stack frames
in order to preserve the context for building a nice error message.
The problem is that `c.inferImplicitValue` is silent.

We solve the problem by keeping error frames in a separate stack.
When building the stack trace we put failed frames back on top.

Also factor out `magnolia:` prefix of error messages.

Fixes #154 (possibly others issues?)
Fixes #178 (second commit)

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -169,7 +169,7 @@ object Magnolia {
       else q"val $name = $rhs"
     }
 
-    def typeclassTree(genericType: Type, typeConstructor: Type): Either[String, Tree] = {
+    def typeclassTree(genericType: Type, typeConstructor: Type, assignedName: TermName): Either[String, Tree] = {
       val searchType = appliedType(typeConstructor, genericType)
       val deferredRef = for (methodName <- stack find searchType) yield {
         val methodAsString = methodName.decodedName.toString
@@ -178,7 +178,7 @@ object Magnolia {
 
       deferredRef.fold {
         val path = ChainedImplicit(s"$prefixName.Typeclass", genericType.toString)
-        val frame = stack.Frame(path, searchType, termNames.EMPTY)
+        val frame = stack.Frame(path, searchType, assignedName)
         stack.recurse(frame, searchType) {
           Option(c.inferImplicitValue(searchType))
             .filterNot(_.isEmpty)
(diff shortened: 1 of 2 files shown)


joroKr21:
@propensive looks like the Travis setup is not working

ghostdogpr:
@joroKr21 I tried your branch on my codebase and not only does it remove the compiler error, but it actually compiles just fine (I was expecting a "nicer" error, but still an error). Does that make sense based on what you changed?

adamw:
@joroKr21 yeah, travis is fixed in PR #177 :)

looks great!

joroKr21:
@ghostdogpr if it works I'd say it's expected :smile: Maybe the second commit matters, as it also fixes a bug.

ghostdogpr:
@joroKr21 a big THANKS then 😄 

@adamw @propensive would it be possible to do a release with this fix?

propensive:
@joroKr21 Thank you so much for working on this! @adamw will do the release, and I'll make an announcement on Twitter a bit later. :)

adamw:
Released in 0.12.0, should be in central soon :)

PR #183, opened 2019-09-30
enable derivation of case class with private constructors

yanns:
https://github.com/propensive/magnolia/issues/138#issuecomment-536532038

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -209,12 +209,6 @@ object Magnolia {
       val isCaseObject = classType.exists(_.isModuleClass)
       val isSealedTrait = classType.exists(_.isSealed)
 
-      val hasPrivateContructor =
-        genericType.decls.collectFirst {
-          case m: MethodSymbol if m.isConstructor =>
-            m.isPrivate
-      }.getOrElse(false)
-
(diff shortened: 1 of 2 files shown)


yanns:
Part of my effort to port our serialization/deserialization libraries to use magnolia: https://github.com/sphereio/sphere-scala-libs/tree/derive_with_magnolia

yanns:
@jto this is a revert of https://github.com/propensive/magnolia/pull/139 to enable derivation where the private constructor is accessible.

If I try to derive the type class instance where the private constructor is not accessible, scalac throws a compilation error:
```
/tests.scala:246:15: constructor PrivateCons in class PrivateCons cannot be accessed in <$anon: magnolia.CaseClass[magnolia.examples.Show.Typeclass,magnolia.tests.PrivateCons]>
[error]       Show.gen[PrivateCons].show(PrivateCons("dada", "phil"))
[error]               ^
```
The compilation error is expected, and I found the error message friendly enough to understand what is going on.

@propensive any thoughts about this PR?

jto:
Merged it since no one commented

aeons:
Could we also have a release with this change? :)

jto:
I don't have the permissions to publish a release. Ping @propensive :)

PR #184, opened 2019-10-22
Add `constructEither` to `CaseClass`

### LINKED ISSUE #181: Accumulating errors in decoder-like type classes
For decoder-like type classes it's often useful to report errors for all failing fields of a case class, not only for the first one. In PureConfig we use an `Either[ConfigReaderFailures, A]` to accumulate failures and I'd like to reproduce this behavior without having to fall back to `rawConstruct`. Am I correct in assuming there is no current workaround for this?

How difficult would it be to provide a `constructApplicative`? One challenge would be to generate instances automatically like Magnolia does with Mercator, but I'd be happy with users having to provide their own instances.

ruippeixotog:
Closes #181.

This PR adds a `constructEither` method that allows users to construct case classes in a type safe manner given how to construct each field, where each field can fail with an error and where errors can be accumulated.

This is a special case of a more general `constructApplicative`, which I chose not to implement because we lack an `Applicative` type class and it's not easy to generate one like we do with `Monad` (see my comment in https://github.com/propensive/magnolia/issues/181#issuecomment-542895033). While less general, this method should still be useful for most purposes where people would use `Applicative`.

I tested this by running the following file in `examples/shared/src/main/scala`:

```scala

object DecoderSafeApp extends App {
  def decodeSafe[A](str: String)(implicit decoder: DecoderSafe[A]): Either[String, A] =
    decoder.decode(str)

  println(decodeSafe[String]("abc"))
  println(decodeSafe[Int]("3"))
  println(decodeSafe[Int]("n"))

  case class MyClass(n: Int, str: String, m: Int)
  println(decodeSafe[MyClass]("(n=3,str=abc,m=1)"))
  println(decodeSafe[MyClass]("(n=n,str=abc,m=1)"))
  println(decodeSafe[MyClass]("(n=n,str=abc,m=m)"))
}
```

The output was, as expected:

```
Right(abc)
Right(3)
Left(illegal number: n)
Right(MyClass(3,abc,1))
Left(illegal number: n)
Left(illegal number: n
illegal number: m)
```

### DIFF
--- a/core/shared/src/main/scala/interface.scala
+++ b/core/shared/src/main/scala/interface.scala
@@ -166,6 +166,8 @@ abstract class CaseClass[Typeclass[_], Type] (
 
   def constructMonadic[Monad[_], PType](makeParam: Param[Typeclass, Type] => Monad[PType])(implicit monadic: Monadic[Monad]): Monad[Type]
 
+  def constructEither[Err, PType](makeParam: Param[Typeclass, Type] => Either[Err, PType]): Either[List[Err], Type]
+
   /** constructs a new instance of the case class type
     *
     *  Like [[construct]] this method is implemented by Magnolia and lets you construct case class
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -251,6 +251,9 @@ object Magnolia {
             def constructMonadic[$f[_], Return](makeParam: _root_.magnolia.Param[$typeConstructor, $genericType] => $f[Return])(implicit monadic: _root_.mercator.Monadic[$f]): $f[$genericType] =
               monadic.point(${genericType.typeSymbol.asClass.module})
 
+            def constructEither[Err, PType](makeParam: _root_.magnolia.Param[$typeConstructor, $genericType] => _root_.scala.Either[Err, PType]): _root_.scala.Either[_root_.scala.List[Err], $genericType] =
+              _root_.scala.Right(${genericType.typeSymbol.asClass.module})
+
             def rawConstruct(fieldValues: _root_.scala.Seq[_root_.scala.Any]): $genericType =
               ${genericType.typeSymbol.asClass.module}
           })
@@ -368,6 +371,31 @@ object Magnolia {
           ) yield new $genericType(..${forParams.map(_._1)})
         """
 
+        val constructEitherImpl = if (caseParams.isEmpty) q"_root_.scala.Right(new $genericType())" else {
+          val eitherVals = caseParams.zipWithIndex.map { case (typeclass, idx) =>
+            val part = TermName(s"p$idx")
+            val pat = TermName(s"v$idx")
+            (
+              part,
+              if(typeclass.repeated) q"$pat: _*" else q"$pat",
+              q"val $part = makeParam($paramsVal($idx)).asInstanceOf[_root_.scala.Either[Err, ${typeclass.paramType}]]",
+              pq"_root_.scala.Right($pat)",
+            )
+          }
+
+          q"""
+           ..${eitherVals.map(_._3)}
+           (..${eitherVals.map(_._1)}) match {
+             case (..${eitherVals.map(_._4)}) =>
+               _root_.scala.Right(new $genericType(..${eitherVals.map(_._2)}))
+             case _ =>
+               _root_.scala.Left(
+                 _root_.scala.List(..${eitherVals.map(_._1)}).filter(_.isLeft).map(_.left.get)
+               )
+           }
+         """
+        }
+
         Some(q"""{
             ..$preAssignments
             val $paramsVal: $scalaPkg.Array[$magnoliaPkg.Param[$typeConstructor, $genericType]] =
@@ -388,6 +416,10 @@ object Magnolia {
 
               def constructMonadic[$f[_], Return](makeParam: _root_.magnolia.Param[$typeConstructor, $genericType] => $f[Return])(implicit monadic: _root_.mercator.Monadic[$f]):$f[$genericType] = {
                 $constructMonadicImpl
+              }
+
+              def constructEither[Err, PType](makeParam: _root_.magnolia.Param[$typeConstructor, $genericType] => _root_.scala.Either[Err, PType]): _root_.scala.Either[_root_.scala.List[Err], $genericType] = {
+                $constructEitherImpl
               }
 
               def rawConstruct(fieldValues: _root_.scala.Seq[_root_.scala.Any]): $genericType = {
--- a/examples/shared/src/main/scala/decodeSafe.scala
+++ b/examples/shared/src/main/scala/decodeSafe.scala
@@ -1,0 +1,88 @@
+/* Magnolia, version 0.10.0. Copyright 2018 Jon Pretty, Propensive Ltd.
+ *
+ * The primary distribution site is: http://co.ntextu.al/
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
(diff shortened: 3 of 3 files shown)


pjrt:
Though we lack the applicative typeclass, we can just add `ap` in mercator.
```
        def ap[A, B](from: Monad[A])(fn: Monad[A => B]): Monad[B] = fn.flatMap(f => from.map(f))
```

And then the user can override `ap` to be more specific. 

However I do think magnolia should just use cats or scalaz instead. 

Either way, I'll accept this as it is too :D

joroKr21:
But `ap` defined via `flatMap` would just short circuit and not accumulate errors.

pjrt:
@jodersky yes, the user would need to override it and make it do so (the user already needs to define `Monadic` for whatever they are using). That definition would just make it so that Mercator can continue to work as expected (since everything has a flatMap, but nothing has an `ap`).

pjrt:
Another thing: in reality, we `constructMonadic` should just use applicative, not monad (assuming we get it). 

If you think about how a case class is built, it is always built as a set of independent requirements. 

```
MyCaseClass <$> p1 <*> p2 <*> p3 ...
```
There really shouldn't be a case where `p2` would depend on `p1`. So, assuming we add `ap` to `Monadic`, `constructMonadic` should just be changed to use `ap` (no new function needed).

ruippeixotog:
I don't really like the idea of adding `ap` to Mercator. From Magnolia's standpoint, an `ap` that chains computations is never the behavior we want. From Mercator's standpoint, it misleads users into thinking that it is somehow able to generate proper implementations. It would also be very confusing as an API for a `Monad` instance to have an `ap` that can be overridden to provide parallel behavior.

It also makes Mercator more complex, to a point where cats or scalaz are better options (currently it limits itself to generating type classes proxying calls directly to methods, making it very lightweight). I'd prefer to think practical here and stay with `constructEither`, which does not bring pure FP complexities into this. @propensive, what's your opinion on this? 

> the user already needs to define `Monadic` for whatever they are using

That's not quite true, `Monadic` is generated through a macro for every class that implements `map`, `flatMap` and has a companion object with an appropriate `apply`, duck-typing style. That makes `constructMonadic` work out-of-the-box for the great majority of users (myself included).

> If you think about how a case class is built, it is always built as a set of independent requirements.

That's true, but you may want the ability to short circuit for performance reasons. If you're not accumulating errors you may as well stop after the first error occurs. If we had `constructApplicative` I'd consider removing `constructMonadic` altogether (but never define monad in terms of ap).

pjrt:
>  From Magnolia's standpoint, an ap that chains computations is never the behavior we want. 

Why do you say that? As my post shows, there's never really a case where you want to have `>>=` when building case classes. In fact, if you look at the generated code magnolia makes for the `for-comp`, it looks like this:

```scala
for { 
  p1 <- mP1
  p2 <- mP2
  p3 <- mP3
 ...
} yield MyCaseClass(p1, p2, p3, ..)
```

If Scala had `-XApplicativeDo`, we can just enable it and boom, we are using `ap` instead of `flatMap` 😄 ([there's a proposal for it on dotty, it seems](https://contributors.scala-lang.org/t/new-syntax-applicative-desugaring-in-for-comprehensions-with-pr/690))

> From Mercator's standpoint, it misleads users into thinking that it is somehow able to generate proper implementations.

Not sure what "proper" means here. And I don't think it will mislead. In Haskell, it is very common to define `>>=` and `<*>` in terms of each other for default implementation. 

> It also makes Mercator more complex, to a point where cats or scalaz are better options

Yeah, I feel it should just do that. This `Monadic` thing is pretty confusing.

> That's not quite true, Monadic is generated through a macro for every class that implements map, flatMap and has a companion object with an appropriate apply, duck-typing style. That makes constructMonadic work out-of-the-box for the great majority of users (myself included).

Unn, then I'm not sure what we are doing wrong, cuz we have made this Monadic thing for everything manually 🤔 

> That's true, but you may want the ability to short circuit for performance reasons. If you're not accumulating errors you may as well stop after the first error occurs. If we had constructApplicative I'd consider removing constructMonadic altogether (but never define monad in terms of ap).

Fair point. Like I said, `constructEither` works too 👍

ruippeixotog:
> Why do you say that? As my post shows, there's never really a case where you want to have >>= when building case classes.

I think we are saying the same thing 😄 I just commented that defining `ap` as `fn.flatMap(f => from.map(f))` is not the behavior we want (even though we could). Having applicative out-of-the-box like the dotty proposal you sent would be the way to go.

> Not sure what "proper" means here. And I don't think it will mislead. In Haskell, it is very common to define >>= and <*> in terms of each other for default implementation.

Bad choice of words, sorry! By "proper" I meant generating an implementation derived directly from Scala methods like it does with `map` and `flatMap`, as opposed to a default implementation.

> Yeah, I feel it should just do that. This Monadic thing is pretty confusing.

I guess that both Magnolia and Mercator are about enabling generic programming by leveraging macros and the provided stdlib as opposed to using 3rd party libs that are heavy (like Shapeless) or lock you in (like cats vs scalaz). I can see how this more "standalone" approach might be attractive, but it also means you can't rely on the power of those more complete libs to do your work 🙂

pjrt:
Built this PR and it fixes my issue. ETA on this getting released?

pjrt:
Pinging again. Is there a reason we can't merge this?

ruippeixotog:
@propensive @adamw could you please take a look at this?

adamw:
I suppose @ghostbuster91 might be a good person to take a look :)

ghostbuster91:
LGTM, although if we could make the `constructEither` implementation more readable, that would be great.

propensive:
@pjrt Sorry for the delay... I haven't had time to stay on top of this. Thanks @adamw and @ghostbuster91 for taking a look.

I have one concern about this which is that I don't know if it has any (compile-time) performance impact, and whether it makes larger derivations impossible. When I was developing Magnolia, I was able to construct examples which (IIRC) hit the bytecode size limit for a method, and that made it desirable to minimize the amount of code that gets generated by the macro. Magnolia was able to do derivations which Shapeless couldn't, due to this size limit.

I don't have any tests for this, and I haven't thought enough about bytecode generation to work out whether it's going to make a significant difference.

So I'm minded to merge this, with a couple of points:
- thank you very much to everyone for contributing, especially when I have so little time for reviewing
- if anyone has the chance to find the limits of the "biggest" derivations we can do, preferably before and after this PR, that would be great
- I like the idea of a `constructApplicative` option as well, and would prefer that more general solution to `constructEither`, but it requires some work in Mercator. I can probably find a bit of time to update Mercator to include a stripped-down `Monadic` type.
- I don't want a dependency on Cats when we only need one (or two) typeclasses. I have a bit of an allergy to unnecessary dependencies, but in this case it hopefully seems quite obvious that it would be bringing in a large surface area of potential conflict with other libraries' dependencies.

So let's merge, and I'll cut a new release, but on the understanding that a more general `constructApplicative` method will replace `constructEither` one day.

propensive:
Version `0.12.3` should be on its way to Maven Central including this PR, if I've done everything correctly. Thanks again, everyone!

propensive:
@ruippeixotog @pjrt I just published a new version of Mercator (version 0.3.0) which should generate applicatives on-demand (called `Applicative[F[_]]`). So `constructApplicative` should be possible now.

pjrt:
@propensive  Awesome, thanks!

ruippeixotog:
@propensive looking at the latest version of Mercator it looks that you are generating functors, not applicatives. In order for it to be an applicative, the [new type class](https://github.com/propensive/mercator/blob/master/core/shared/src/main/scala/mercator.scala#L100-L104) should have an additional `def ap[A, B](from: F[A])(fn: F[A => B]): F[B]` method. To my knowledge, there is no easy way to generate it from standard Scala collection methods, which is why I chose not to add it to Mercator in the first place.

pjrt:
Been 6 hours now, but it seems 0.12.3 didn't make it to maven yet :thinking:

propensive:
Yes, of course I am... whoops. ;)

propensive:
@pjrt It *looks* like it was successful:
https://travis-ci.org/propensive/magnolia

pjrt:
Unn, sonatype must be behind then, it isn't here as of this writing: https://mvnrepository.com/artifact/com.propensive/magnolia

propensive:
It's here, though: https://repo.maven.apache.org/maven2/com/propensive/magnolia_2.12/0.12.3/

pjrt:
Welp, I'm an idiot. The issue is on my side. Thanks! 👍

PR #192, opened 2019-12-05
Use `collect` instead of `left.get`

iRevive:
This PR closes https://github.com/propensive/magnolia/issues/191.

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -406,7 +406,7 @@ object Magnolia {
                _root_.scala.Right(new $genericType(..${eitherVals.map(_._2)}))
              case _ =>
                _root_.scala.Left(
-                 _root_.scala.List(..${eitherVals.map(_._1)}).filter(_.isLeft).map(_.left.get)
+                 $magnoliaPkg.Magnolia.keepLeft(..${eitherVals.map(_._1)})
                )
            }
          """
@@ -602,6 +602,10 @@ object Magnolia {
       val msg = "`" + typeName + "` has " + paramsLength + " fields, not " + fieldValues.size
       throw new java.lang.IllegalArgumentException(msg)
     }
+
+  final def keepLeft[A, B](values: Either[A, B]*): List[A] =
+    values.toList.collect { case Left(v) => v }
+
 }
 
 private[magnolia] final case class DirectlyReentrantException()


iRevive:
Well, this exercise is not that trivial that I thought. 
I ran into the compiler bug https://github.com/scala/bug/issues/5465. 

I tried to define a case statement using trees API instead of quasiquotes:
```scala
val matcher = CaseDef(
  Apply(
    Select(Select(Ident(termNames.ROOTPKG), TermName("scala")), TermName("Left")),
    List(Ident(TermName("v")))
  ),
  EmptyTree,
  Ident(TermName("v"))
)

q"""
...
_root_.scala.Left(
  _root_.scala.List(..${eitherVals.map(_._1)}).collect({ case $matcher })
)
"""
```

But this implementation fails with the following error:
```not found: value v```

Is there is a way to overcome this compiler limitation?

**Comment by joroKr21** [comment]
How about adding a protected helper method in `CaseClass`?
It would also reduce the size of the generated bytecode.

**Comment by iRevive** [comment]
@joroKr21  thanks, the helper method solved the issue.

**Comment by aeons** [comment]
Is this ready to be merged? It triggers a deprecation warning in 2.13 as well.

**Comment by markus1189** [comment]
Just came across this issue.  We are using `fatal-warnings` and this leads to an error due to a library that uses magnolia underneath.

Would be cool to have this merged and released

File: core/shared/src/main/scala/magnolia.scala [hunk 1]
```diff
@@ -406,7 +406,7 @@ object Magnolia {
                _root_.scala.Right(new $genericType(..${eitherVals.map(_._2)}))
              case _ =>
                _root_.scala.Left(
-                 _root_.scala.List(..${eitherVals.map(_._1)}).filter(_.isLeft).map(_.left.get)
```
adamw: can't we simply do `_root_.scala.List(..${eitherVals.map(_._1)}).collect { case Left(v) => v }`? seems to work locally :)

iRevive: That was the first thing I tried. Unfortunately, it does not work with Scala 2.12. 
Build example: https://travis-ci.org/propensive/magnolia/jobs/621026977?utm_medium=notification&utm_source=github_status

adamw: Ah, I didn't clean before compile locally :) you're right, it fails

adamw:
Should be released in 0.12.4, thanks @propensive !

adamw:
& @iRevive of course :)

PR #194, opened 2019-12-05
Fix #193: Tuple 22 issue

### LINKED ISSUE #193: BUG: "Tuples with 23 arity aren't supported" error on 0.12.3
0.12.3 (the contructEither change, https://github.com/propensive/magnolia/pull/184) seems to have broken the derivation for very long case classes (over 22 parameters). 

See here for a patch that adds a test that shows this: [patch.txt](https://github.com/propensive/magnolia/files/3927299/patch.txt)

I can't reproduce this on 0.12.2 or before it.

pjrt:
The issue was related to how we combine the Eithers into the target type. We were building a tuple the size of the case class, but tuples have limits that case classes don't have (anymore). 

So my solution is to limit the size of the tuples we do. If we had `ap` and curried functions, this wouldn't be an issue, but we don't have that atm.

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -387,22 +387,29 @@ object Magnolia {
           ) yield new $genericType(..${forParams.map(_._1)})
         """
 
-        val constructEitherImpl = if (caseParams.isEmpty) q"_root_.scala.Right(new $genericType())" else {
-          val eitherVals = caseParams.zipWithIndex.map { case (typeclass, idx) =>
-            val part = TermName(s"p$idx")
-            val pat = TermName(s"v$idx")
-            (
-              part,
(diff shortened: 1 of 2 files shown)


pjrt:
@propensive can you take a quick look at this? 0.12.3 is currently broken due to this.

pjrt:
Been 7 days...can we get this in? it is making 0.12.3 impossible to use for us.

danielkarch:
Hi, I would also be happy to see this merged. I am trying to update to the newest tapir version but just learned that we have some case classes with a _lot_ of parameters ...

0lejk4:
+1, need this for tapir too.

0lejk4:
@propensive Please can you look at this PR :cry:

propensive:
Sorry for ignoring everyone for so long! It's going into a release any minute now...

0lejk4:
Huge thanks @propensive :+1:

PR #196, opened 2019-12-16
Sort subtypes to ensure consistent ordering (#195)

### LINKED ISSUE #195: SealedTraits are not listed in stable order
Currently the ordering of sealed trait types is based on `ClassSymbol.knownDirectSubclasses.toList` (see [src/main/scala/magnolia.scala#L129](https://github.com/propensive/magnolia/blob/24e18384666b60eba9119c9de00f97afc4d129cb/core/shared/src/main/scala/magnolia.scala#L129)).

`knownDirectSubclasses` is a `Set` with unspecified ordering (dependent on hash algorithms and the order in which the compiler visits types), and so the results of `toList`-ing it can change unexpectedly (i.e. can change even if the order they are declared in the code remains unchanged).

How much this is a problem obviously depends on the use case, but imho it would be better to expose these in a stable order. Suggest alphabetically sorted based on fullName, this would have the benefit of being stable against reordering the code and would match the logic in [Shapeless Generic](https://github.com/milessabin/shapeless/blob/ab5a1a6bb1e01e13fd417e0ee0d8726d54344d84/core/src/main/scala/shapeless/generic.scala#L423).

jfwilson:
This PR closes #195.

NB I reused `TypeNameInfo` to test the order of sealed traits but happy to introduce a new type for the test if you prefer. I also extended the `Color` trait to have more than 4 subtypes to demonstrate the unknown order problem. Not many comments in the code so didn't add one but happy to put one in if you think it would be helpful.

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -125,10 +125,10 @@ object Magnolia {
       .filter(encl => encl.isVal || encl.isLazy)
       .toSet[Symbol]
 
-    def knownSubclasses(sym: ClassSymbol): List[Symbol] = {
-      val children = sym.knownDirectSubclasses.toList
+    def knownSubclasses(sym: ClassSymbol): Set[Symbol] = {
+      val children = sym.knownDirectSubclasses
       val (abstractTypes, concreteTypes) = children.partition(_.isAbstract)
-      abstractTypes.map(_.asClass).flatMap(knownSubclasses(_)) ::: concreteTypes
+      abstractTypes.map(_.asClass).flatMap(knownSubclasses(_)) ++ concreteTypes
     }
 
     def annotationsOf(symbol: Symbol): List[Tree] = {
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/magnolia.scala
```diff
@@ -446,7 +446,7 @@ object Magnolia {
           }""")
       } else if (isSealedTrait) {
         checkMethod("dispatch", "sealed traits", "SealedTrait[Typeclass, _]")
-        val genericSubtypes = knownSubclasses(classType.get)
+        val genericSubtypes = knownSubclasses(classType.get).toList.sortBy(_.fullName)
```
joroKr21: Hmm, does `fullName` include the package? In most cases this would be the same for all subclasses, i.e. redundant.

jfwilson: Yes, it includes package name. I realise that in most cases owner name is not relevant but this is a library that is to be used by all sorts of different projects and is entirely possible for projects to nest classes, e.g.
```
sealed trait Animal
object Domestic {
    case object Cat extends Animal
}
object Wild {
    case object Cat extends Animal
}
```
As mentioned in the corresponding issue I think there is some benefit in having ordering consistent with the way shapeless generic does it, but if you would prefer to sort by Symbol name alone that's fine by me.

joroKr21: Ah I didn't realize shapeless used `fullName`. Nvm then I think it's good.

PR #197, opened 2019-12-18
Remove usage of Either.LeftProjection.get

markus1189:
In Scala 2.13 this method was deprecated, and let's projects that use
magnolia fail if fatal warnings are active.

I replaced the `filter` followed by `.left.get` with a `collect` to
get rid of the warning.

For reference, the exact message is:

```scala
method get in class LeftProjection is deprecated (since 2.13.0): use `Either.swap.getOrElse` instead
```


markus1189:
Oops, just saw #192 , closing

PR #199, opened 2019-12-19
Do not infer the type of the underlying eithers

iRevive:
This PR closes https://github.com/propensive/magnolia/issues/198.


pjrt:
Just hit this issue. Do we have an ETA on when it can be released?

PR #200, opened 2019-12-19
Fix tests

joroKr21:
The problem is that sbt succeeds even when the tests fail,
so we don't notice.


propensive:
Thanks, @joroKr21—well spotted!

PR #207, opened 2020-02-22
Replace usage of javax annotation

joroKr21:
`javax.annotation` is not part of the JDK since Java 11.


propensive:
Thanks for the quick fix @joroKr21 and the reviews, @xuwei-k & @regadas!

Issue #216, opened 2020-03-25
Runtime StackOverflowError with nested sealed hierarchy


einholen:
Hi! I'm _intermittently_ getting a `StackOverflowError` at runtime because the Magnolia macro resolves itself implicitly instead of the actual concrete type class.
I saw #66 and #190 but they seem to be different, so creating a new issue. Feel free to close if it's a dupe and unresolvable in Scala 2 🙂 

So there's a sealed trait with a bunch of sealed subtraits each of which has a bunch of concrete case classes underneath. We use magnolia to dispatch specific type classes based on an instance of the sealed trait at runtime. We don't use `combine`, only `dispatch`. All branches of the hierarchy terminate, and each concrete class has an implicit defined that Magnolia is supposed to find.

With a hierarchy like this
```scala
sealed trait Base
sealed trait Sub1 extends Base { def param: Int }
case class Concrete1(param: Int) extends Sub1
sealed trait Sub2 extends Base { def param: String }
case class Concrete2(param: String) extends Sub2
```

and implicits for concrete case classes defined in a separate compilation unit from the Magnolia macro

```scala
object Show {
  type Typeclass[T] = Show[T]
  def combine[T <: Base](caseClass: CaseClass[Show, T]): Show[T] = throw new UnsupportedOperationException
  def dispatch[T <: Base](sealedTrait: SealedTrait[Show, T]): Show[T] = new Show[T] {
    def show(value: T): String = sealedTrait.dispatch(value) { subtype =>
      subtype.typeclass.show(subtype.cast(value))
    }
  }
  implicit def gen[T <: Base]: Show[T] = macro Magnolia.gen[T]
}

object Implicits {
  implicit val showConcrete1: Show[Concrete1] = (c: Concrete1) => s"Concrete1:${c.param}"
  implicit val showConcrete2: Show[Concrete2] = (c: Concrete2) => s"Concrete2:${c.param}"
}
``` 

Calling `implicitly[Show[Base]].show(Concrete2("hello"))` _sometimes_ results in a runtime `StackOverflowError`.

After a clean compilation the error is gone, but changes to adjacent code seem to sometimes resurrect it.
What I'm seeing in debug inside `sealedTrait.dispatch` is, `subtypesArray` contains `Concrete1` and `Sub2`. A value of `Concrete2` thus matches `Sub2`, and then `subtype.typeclass` resolves to `Show.gen` again and again instead of recursing inside the sealed trait and finding the actual concrete implicit.
[...]

PR #217, opened 2020-03-29
Use ThreadLocal stack and workset

joroKr21:
Hoping that this will solve issues when using the Hydra compiler.

Ref #216 but I'm not sure if it actually fixes the problem.

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -706,15 +706,17 @@ private[magnolia] object CompileTimeState {
   object Stack {
     // Cheating to satisfy Singleton bound (which improves type inference).
     private val dummyContext: whitebox.Context = null
-    private val global = new Stack[dummyContext.type]
-    private val workSet = mutable.Set.empty[whitebox.Context#Symbol]
+    private val threadLocalStack = ThreadLocal.withInitial[Stack[dummyContext.type]](() => new Stack[dummyContext.type])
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/magnolia.scala
```diff
@@ -706,15 +706,17 @@ private[magnolia] object CompileTimeState {
   object Stack {
     // Cheating to satisfy Singleton bound (which improves type inference).
     private val dummyContext: whitebox.Context = null
-    private val global = new Stack[dummyContext.type]
-    private val workSet = mutable.Set.empty[whitebox.Context#Symbol]
+    private val threadLocalStack = ThreadLocal.withInitial[Stack[dummyContext.type]](() => new Stack[dummyContext.type])
+    private val threadLocalWorkSet = ThreadLocal.withInitial[mutable.Set[whitebox.Context#Symbol]](() => mutable.Set.empty)
```
einholen: I haven't looked into magnolia internals nearly enough, so I have zero context - forgive me if it's a stupid question 😄 

I assume no merging is required for different instances of this `ThreadLocal` because everything from the same compilation unit should end up in the same stack, right?

joroKr21: That would be my assumption but it really depends on Hydra.
I would not expect it to switch threads for a single macro expandsion.
But on the other hand if it's using `Future` this won't help.

PR #218, opened 2020-03-29
Fix derivation for case classes with partially private fields

### LINKED ISSUE #215: Magnolia fails to derive instances for a case class with partially private fields
Example:

```scala
final case class Abc(
  private val a: Int,
  private val b: Long,
  c: String
)
object Abc {
  val reader: pureconfig.ConfigReader[Abc] = pureconfig.module.magnolia.semiauto.reader.deriveReader[Abc]
}
```

Expected: 
Success

Actual:
Error:
```
Error:(22, 90) type mismatch;
 found   : String
 required: Int
  val reader: ConfigReader[Abc] = pureconfig.module.magnolia.semiauto.reader.deriveReader[Abc]
````

Original issue in `pureconfig-magnolia`: https://github.com/pureconfig/pureconfig/issues/701

joroKr21:
Use sorted declarations of fields instead of getters.
Synthetic accessors might appear in arbitrary order.
Also remove unnecessary synthetic accessor name regex.

Fixes #215

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -24,8 +24,6 @@ import mercator._
 object Magnolia {
   import CompileTimeState._
 
-  private val SyntheticAccessorName = "(.+)\\$access\\$\\d+".r
-
   /** derives a generic typeclass instance for the type `T`
     *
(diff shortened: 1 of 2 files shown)


propensive:
Looks good, thank you, @joroKr21!

PR #220, opened 2020-03-29
Update dependencies

joroKr21:
Couldn't update mercator though:

```scala
[error] magnolia/examples/shared/src/main/scala/default.scala:31:45: could not find implicit value for parameter monadic: mercator.Monadic[[+B]scala.util.Either[String,B]]
[error]     def defaultValue = ctx.constructMonadic { param =>
```

  * sbt 1.3.8
  * Scala 2.12.11
  * Scala.js 1.0.1
  * xml.bind

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: scala
 jdk:
 - openjdk8
 scala:
-- 2.12.10
+- 2.12.11
 - 2.13.1
 before_install:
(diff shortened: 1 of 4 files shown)


propensive:
I just restarted the build, seeing as it failed.

I'll see if I can get it to use a more recent version of Mercator if I get time this coming week.

Issue #107, opened 2018-06-15
Separate exported instance type and internal dispatch/combine type


vpavkin:
This is a more principled version of #89. I found at least two cases when the typeclass intance being derived would have a type different from the typeclass used for `dispatch` and `combine`.

### 1) `ObjectEncoder[A]`/`Encoder[A]`

In circe, generic derivation provides not just an `Encoder[A]`, but a more specific `ObjectEncoder[A]`, which means that the serialization result is always going to be a json object. But for all the inner steps of the derivation, a more general `Encoder[A]` typeclass is looked for and used. This makes sense - object internals don't have to be objects themselves.

### 2) `Exported[Decoder[A]]`/`Decoder[A]`

This is how prioritization of derived instances works in circe. `auto` derives an instance of `Exported[Decoder[A]]` ([code](https://github.com/circe/circe/blob/master/modules/generic/shared/src/main/scala/io/circe/generic/AutoDerivation.scala#L18)). And there's a [low-priority implicit conversion](https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Encoder.scala#L470) `Exported[Decoder[A]] => Decoder[A]`, which allows the default instances to be selected before the derived ones.

But of course, under the hood derivation of `Exported[Decoder[A]]` combines and dispatches on just `Decoder[A]` (otherwise default instances would not be picked up).

### Possible Solution

I feel that this is not specific to circe and similar concerns might arise in other contexts.

I understand this would complicate things quite a bit, but ideally for proper circe auto derivation we'd need something like this:

```scala
object MagnoliaDecoder {

  type ExportedTypeclass[T] = Exported[Decoder[T]]
  type InternalLookup[T] = Decoder[T]

  // run outer level combine for the exported instance
  // ExportedCaseClass is the same as CaseClass, but parametersArray 
  // will already have InternalLookup instances 
  def combineExported[T](
    caseClass: ExportedCaseClass[ExportedTypeclass, InternalLookup, T]): ExportedTypeclass[T] = ???

  // same idea for the dispatch on the outer level instance
  def dispatchExported[T](
    sealedTrait: ExportedSealedTrait[ExportedTypeclass, InternalLookup, T]): ExportedTypeclass[T] = ???

  // all the internal derivation works in the same way as before - using InternalLookup

  def combine[T](caseClass:    CaseClass[InternalLookup, T]): InternalLookup[T] = ???
[...]

PR #221, opened 2020-03-29
Remove unnecessary condition for lazy val

### LINKED ISSUE #214: lazy val leads to illegal premature access error.
This could be related to the following closed issue (https://github.com/propensive/magnolia/issues/95)

I've came up with code example which partially based on `magnolia`'s examples. Full showcase repo could be found [here](https://github.com/VladPodilnyk/magnolia-exper).

Based o such trait and derivation rules:
```scala
trait Decoder[T] {
  def decode(value: String): T
}

object Decoder {
  /** decodes strings */
  implicit val string: Decoder[String] = (s: String) => s

  /** decodes ints */
  implicit val int: Decoder[Int] = _.toInt

  def gen[T]: Decoder[T] = macro Magnolia.gen[T]

  type Typeclass[T] = Decoder[T]

  def combine[T](ctx: CaseClass[Decoder, T]): Decoder[T] = value => {
    val (_, values) = parse(value)
    ctx.construct {
      param =>
        param.typeclass.decode(values(param.label))
    }
  }

  def dispatch[T](ctx: SealedTrait[Decoder, T]): Decoder[T] = param => {
    val (name, _) = parse(param)
    val subtype   = ctx.subtypes.find(_.typeName.full == name).get
    subtype.typeclass.decode(param)
  }

  private def parse(value: String): (String, Map[String, String]) = ???
}
```
I am trying to provide a convenient type `WithDecoder` which contains `implicit val dec: Decoder[T]`. I've described `WithDecoder` as follows
```scala
abstract class WithDecoder[T]()(implicit decoder: MaterializedDecoder[T]) {
  implicit val dec: Decoder[T] = decoder.value
}

final class MaterializeDerivationMacros(val c: blackbox.Context) {
  import c.universe._

  def materializer[T: c.WeakTypeTag]: c.Expr[MaterializedDecoder[T]] = c.Expr[MaterializedDecoder[T]] {
    q"{ new ${weakTypeOf[MaterializedDecoder[T]]}(_root_.issue.Decoder.gen[${weakTypeOf[T]}]) }"
  }
}

final case class MaterializedDecoder[T](value: Decoder[T])
object MaterializedDecoder {
  implicit def materialize[T]: MaterializedDecoder[T] = macro MaterializeDerivationMacros.materializer[T]
}
``` 

When I've tried to compile the following code snippet, compile throws an error
```scala
object models {
  case class Test(s: String)
  object Test extends WithDecoder[Test]
}
```

```
Error:(5, 23) Implementation restriction: access of lazy value paramTypeclass$macro$3$1 in object Test from <$anon: Function0>, would require illegal premature access to object Test in object models
  object Test extends WithDecoder[Test]
```

So, the question is there any way to fix this?
Could somebody help with that?

joroKr21:
I don't see why semiauto should always be lazy.

Fixes #214

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -184,7 +184,7 @@ object Magnolia {
         case tree => enclosingVals.contains(tree.symbol)
       }
 
-      if (!fullauto || shouldBeLazy) q"lazy val $name: $tpe = $rhs"
+      if (shouldBeLazy) q"lazy val $name: $tpe = $rhs"
       else q"val $name = $rhs"
     }
 


kubukoz:
Apparently, this change actually fixes (part of?) #107, at least if the field's instance is a supertype of the derived instance.

```scala
package magnolia.examples

import language.experimental.macros

import magnolia._

trait Encoder[A] {
  def encode: Any
}

object Encoder {
  implicit val stringEncoder: Encoder[String] = new Encoder[String] {
    def encode: Any = ()
  }

  implicit def listEncoder[T: Encoder]: Encoder[List[T]] = new Encoder[List[T]] {
    def encode: Any = 42
  }
}

trait RootEncoder[A] extends Encoder[A] {
  def encode: String
}

object RootEncoder {
  type Typeclass[A] = Encoder[A]

  def combine[T](ctx: CaseClass[Encoder, T]): RootEncoder[T] = new RootEncoder[T] {
    val encode: String = "root"
  }

  def derive[T]: RootEncoder[T] = macro Magnolia.gen[T]
}

final case class MyType(s: String /* , others: List[MyType] */ )

object MyType {
  val encoder: RootEncoder[MyType] = RootEncoder.derive
}

object demo extends App {
  println(MyType.encoder.encode)
  println(MyType.encoder.encode.length())
}

```

This works after the change in this PR, didn't work before.

I saw no problems with recursive types (uncomment the parameter, make the instance of MyType implicit and it still works).

PR #224, opened 2020-04-04
(Read-only) typeclass support for case class with private constructors

jatcwang:
Many typeclass only require reading fields from their input and do not need to construct any values (e.g. Json Encoders, Show, Eq). Previously, Magnolia couldn't derive instances where the case class constructor isn't accessible. This is due to needing to implement methods in `CaseClass` that construct values.
 
This patch introduces a way for users to signal that they do not need to construct any values for their type class. To opt in, user change the parameter type of the `combine` method they have to implement from `CaseClass` to `ReadOnlyCaseClass`.

```diff
+ def combine[T](ctx: ReadOnlyCaseClass[Eq, T]): Eq[T] = new Eq[T] {
- def combine[T](ctx: CaseClass[Eq, T]): Eq[T] = new Eq[T] {
  ...
}
```
 
It seems like it has the happy side-effect of reducing the compile time of read-only type class derivation by roughly 30% on `GraalVM CE 20.0.0`. (from my totally scientific method of running `bench/clean ; bench/compile` many many times)
I have added a `bench` module in a separate commit you can cherry-pick and compare yourself :)

[ ] Remove `fixme`s
[ ] Better name for ReadOnlyParam?


core/shared/src/main/scala/magnolia.scala
```diff
@@ -326,76 +346,93 @@ object Magnolia {
...
+              _root_.magnolia.CallByNeed($ref),
+              $scalaPkg.Array($annList)
+            )"""
+          }
+        } else {
+          val defaults = headParamList.fold(List[Tree](q"$scalaPkg.None")) { params =>
+            val x = TermName(c.freshName("x"))
+            val A = TypeName(c.freshName("A"))
+            val dummyArgs = params.filterNot(_.isParamWithDefault).map(p => q"${p.name} = $x")
+            // `A <: Nothing` to avoid dead code warnings
```
propensive, line 368: Really? That works?!

core/shared/src/main/scala/magnolia.scala
```diff
@@ -326,76 +346,93 @@ object Magnolia {
...
+
+          caseParams.zip(defaults).zip(annotations).zipWithIndex.map {
+            case (((CaseParam(paramName, repeated, _, paramType, ref), defaultVal), annList), idx) =>
+              val call = if (isValueClass) q"$magnoliaPkg.Magnolia.valueParam" else q"$magnoliaPkg.Magnolia.param"
+              q"""$paramsVal($idx) = $call[$typeConstructor, $genericType, $paramType](
               ${paramName.toString.trim},
               ${if (isValueClass) q"_.$paramName" else q"$idx"},
               $repeated,
               _root_.magnolia.CallByNeed($ref),
               _root_.magnolia.CallByNeed($defaultVal),
```
propensive, line 385: Should we use `$magnoliaPkg` instead of `_root_.magnolia`?

propensive: Whoops... I just noticed this wasn't even one of your additions! ;)

core/shared/src/main/scala/magnolia.scala
```diff
@@ -406,38 +443,49 @@ object Magnolia {
...
+                $constructMonadicImpl
+              }
+             """,
+            q"""
+              def constructEither[Err, PType](makeParam: _root_.magnolia.Param[$typeConstructor, $genericType] => _root_.scala.Either[Err, PType]): _root_.scala.Either[_root_.scala.List[Err], $genericType] = {
+                $constructEitherImpl
+              }
+             """,
+            q"""
+              def rawConstruct(fieldValues: _root_.scala.Seq[_root_.scala.Any]): $genericType = {
```
propensive, line 464: Ok, here we can use `$scalaPkg` and `$magnoliaPkg` in a few places.

propensive:
I'm super impressed with this, @jatcwang! Thank you so much! The 30% performance improvement is particularly exciting. Hopefully it will translate to a couple of seconds saved for compiling Fury...

Do you want to make the tiny changes I mentioned, and then I'll merge it. :)

core/shared/src/main/scala/interface.scala
```diff
@@ -140,12 +169,25 @@ trait Param[Typeclass[_], Type] extends Serializable {
...
+  override val isValueClass: Boolean,
   parametersArray: Array[Param[Typeclass, Type]],
   annotationsArray: Array[Any]
-) extends Serializable {
+) extends ReadOnlyCaseClass[Typeclass, Type](
+  typeName,
+  isObject,
+  isValueClass,
+  // Safe to cast as we're never mutating the array
+  parametersArray.asInstanceOf[Array[ReadOnlyParam[Typeclass, Type]]],
```
joroKr21, line 182: Hmm this made me realise that we are converting to a `Seq` on every method call and I think this is **not** efficient on Scala 2.13 because it does copy on write. I think we should change that (maybe in another PR). @propensive WDYT?

core/shared/src/main/scala/magnolia.scala
joroKr21: Yep 😄 This was just moved though, I did this in #219 and although it still hacky it's less so than what we had before.

core/shared/src/main/scala/interface.scala
```diff
@@ -123,9 +116,45 @@ trait Param[Typeclass[_], Type] extends Serializable {
...
+  *  @tparam Type       generic type of this parameter */
+trait Param[Typeclass[_], Type] extends ReadOnlyParam[Typeclass, Type] {
+
+  /** provides the default value for this parameter, as defined in the case class constructor */
+  def default: Option[PType]
+
   override def toString: String = s"Param($label)"
 }
 
+// FIXME: docs
```
joroKr21, line 134: I'd say either fix it or remove it 😄

core/shared/src/main/scala/interface.scala
propensive: Hmm. I actually have no idea what changed in Scala 2.13's treatment of arrays... What would the change be?

core/shared/src/main/scala/interface.scala
propensive: Removal is fine by me. I'm very much of the opinion that documentation shouldn't hold up code.

core/shared/src/main/scala/interface.scala
joroKr21: Right now if you try to compile it warns:
```scala
[warn] ../magnolia/core/shared/src/main/scala/interface.scala:124:37: method copyArrayToImmutableIndexedSeq in class LowPriorityImplicits2 is deprecated (since 2.13.0): Implicit conversions from Array to immutable.IndexedSeq are implemented by copying; Use the more efficient non-copying ArraySeq.unsafeWrapArray or an explicit toIndexedSeq call
[warn]   final def annotations: Seq[Any] = annotationsArray
```

core/shared/src/main/scala/interface.scala
jatcwang: My own convention is that fixme's are to be dealt with in the PR so I'll fix this up :)

propensive:
Wonderful. Thank you so much! :D

propensive:
I've published version 0.14.0. Not sure if it's available yet, but it's not listed in the Maven Central Search directory...
https://travis-ci.org/github/propensive/magnolia?utm_medium=notification&utm_source=github_status

ghostdogpr:
@jatcwang @joroKr21 I've just tried using this and it seems I am losing the annotations on parameters (only on parameters, seems okay on the case classes themselves). I looked at it in debug and it seems that `ReadOnlyParam#annotations` returns an **array of array** of annotations (instead of an array of annotations).

joroKr21:
Yes, seems like a bug. Fix incoming.

jatcwang:
Sorry! Definitely should've added more tests. (I don't remember touching `annList` at all though..!)

ghostdogpr:
Thanks for the quick fix! All my tests are green with the new version 😄

Issue #228, opened 2020-04-09
magnolia.gen fails if subtype specifies type parameter


jfwilson:
Have encountered the below error dealing with parameterised types. For example this code:
```scala
sealed trait TypedTree[A]

object TypedTree {
  case class Branch[A](left: TypedTree[A], right: TypedTree[A]) extends TypedTree[A]

  case class StringLeaf(value: String) extends TypedTree[String]

  case class IntLeaf(value: Int) extends TypedTree[Int]
}

val stringTree: TypedTree[String] = TypedTree.Branch(TypedTree.StringLeaf("LHS"), TypedTree.StringLeaf("RHS"))
Show.gen[TypedTree[String]].show(stringTree)
```
fails `Show` generation with
```
type arguments [magnolia.examples.Show.Typeclass,TypedTree[String],TypedTree.IntLeaf] do not conform to method subtype's type parameter bounds [Tc[_],T,S <: T]
```

If I remove `case class IntLeaf(value: Int) extends TypedTree[Int]` then it works for `String` but still fails for `Int`.

What I would like is for the type parameter to be taken into account, i.e. if I am generating a `Typeclass[TypedTree[String]]` then only subtypes that can possibly match `TypedTree[String]` are considered. I am not sure of the details but shapeless manages to achieve this.

propensive:
This is interesting. It's quite a simple example (thank you!) but it seems a little strange from a modelling point of view... but that's fine. ;)

It should be possible for the macro to filter the set of subtypes to check that they conform to the overall type, but I haven't thought through (plus, I don't remember enough about it!) the implications for a recursive type like `TypedTree`. I think it wouldn't be too hard to try it, and see if it works, but I'd like to include some tests for a covariant typeclass derivation too to see how that behaves.

PR #229, opened 2020-04-09
Exclude invalid subtypes (#228)

jfwilson:
As far as I can tell adding an `applied <:< genericType` subtype check fixes issue #228.


propensive:
@jfwilson Wow, you implemented my suggestion even momentarily before I suggested it!

propensive:
Thank you for the contribution, @jfwilson! I'll publish a new version now.

propensive:
Version 0.14.2 should be in Maven Central soon.

PR #230, opened 2020-04-09
Fix default parameters computation

### LINKED ISSUE #227: Two parameters with default values cause type mismatch error
In version 0.14.1, typeclass derivation fails if a case class has two parameters with default values:

```scala
import org.scalatest.funsuite.AnyFunSuite

case class ParamsWithDefault(a: Int = 3, b: Int = 4)

class DefaultValuesTest extends AnyFunSuite {

  test("default values") {
    MyTypeClass.gen[ParamsWithDefault]
  }
}
```

Compilation error:

```
Error:(10, 20) type mismatch;
 found   : issues.ParamsWithDefault.type
 required: Int
    MyTypeClass.gen[ParamsWithDefault]
```

The code compiles with only one default value. Full typeclass code:

```scala

import magnolia._

trait MyTypeClass[T] {
  def name: String
}

object MyTypeClass {

  implicit def gen[T]: Typeclass[T] = macro Magnolia.gen[T]

  type Typeclass[T] = MyTypeClass[T]

  implicit val forInt: Typeclass[Int] = new Typeclass[Int] {
    def name: String = "Int"
  }

  def dispatch[T](ctx: SealedTrait[Typeclass, T]): Typeclass[T] =
    new Typeclass[T] {
      def name: String = s"sealed trait ${ctx.typeName.full}"
    }

  def combine[T](ctx: CaseClass[Typeclass, T]): Typeclass[T] =
    new Typeclass[T] {
      def name: String = s"case class ${ctx.typeName.full}"
    }

}
```

joroKr21:
Fixes #227

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -362,16 +362,17 @@ object Magnolia {
             )"""
           }
         } else {
-          val defaults = headParamList.fold(List[Tree](q"$scalaPkg.None")) { params =>
-            val x = TermName(c.freshName("x"))
-            val A = TypeName(c.freshName("A"))
-            val dummyArgs = params.filterNot(_.isParamWithDefault).map(p => q"${p.name} = $x")
-            // `A <: Nothing` to avoid dead code warnings
-            val dummy = c.typecheck(q"def $x[$A <: Nothing]($x: $A) = new $genericType(..$dummyArgs)")
-            val defaultParams = dummy.collect { case sel: Select if sel.symbol.isSynthetic => sel }.iterator
-            for (p <- params) yield {
-              if (!p.isParamWithDefault) q"$scalaPkg.None"
-              else q"$scalaPkg.Some(${defaultParams.next()})"
+          val defaults = headParamList.fold[List[Tree]](Nil) { params =>
+            if (params.exists(_.isParamWithDefault)) {
+              val x = TermName(c.freshName("x"))
+              val A = TypeName(c.freshName("A"))
+              val dummyArgs = params.filterNot(_.isParamWithDefault).map(p => q"${p.name} = $x")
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/magnolia.scala
```diff
@@ -362,16 +362,17 @@ object Magnolia {
...
-              if (!p.isParamWithDefault) q"$scalaPkg.None"
-              else q"$scalaPkg.Some(${defaultParams.next()})"
+          val defaults = headParamList.fold[List[Tree]](Nil) { params =>
+            if (params.exists(_.isParamWithDefault)) {
+              val x = TermName(c.freshName("x"))
+              val A = TypeName(c.freshName("A"))
+              val dummyArgs = params.filterNot(_.isParamWithDefault).map(p => q"${p.name} = $x")
+              // `A <: Nothing` to avoid dead code warnings
+              val dummy = c.typecheck(q"def $x[$A <: $scalaPkg.Nothing]($x: $A) = new $genericType(..$dummyArgs)")
+              val defaults = dummy.collect { case sel @ Select(_, name) if name.toString.contains("$default$") => sel }.iterator
```
joroKr21, line 372: If you wonder why we can't just get the parameters in a list, there is this:
```scala
def x$macro$258[A$macro$259 <: scala.Nothing](x$macro$258: A$macro$259) = {
  val x$2: A$macro$259 = x$macro$258;
  val x$3: A$macro$259 = x$macro$258;
  val x$4: Int = magnolia.tests.Item.`<init>$default$2`;
  new magnolia.tests.Item(x$2, x$4, x$3)
}
```
The compiler rewrote it to preserve execution order when there is a default parameter in the middle.

propensive: That's crazy. But we're playing with fire, anyway.

PR #232, opened 2020-04-15
Fix annotations of ReadOnlyParam by eliminating code duplication

joroKr21:
Bug was introduced in #224, we missed it during review.

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -311,13 +311,25 @@ object Magnolia {
           else { case p: TermSymbol if p.isCaseAccessor && !p.isMethod => p }
         )
 
-        case class CaseParam(
-          paramName: TermName,
-          repeated: Boolean,
-          typeclass: Tree,
-          paramType: Type,
-          ref: TermName
-        )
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/magnolia.scala
```diff
@@ -352,18 +364,8 @@ object Magnolia {
...
-          caseParams.zip(annotations).zipWithIndex.map {
-            case ((CaseParam(paramName, repeated, _, paramType, ref), annList), idx) =>
-            val call = if (isValueClass) q"$magnoliaPkg.Magnolia.readOnlyValueParam"
-                       else q"$magnoliaPkg.Magnolia.readOnlyParam"
-            q"""$paramsVal($idx) = $call[$typeConstructor, $genericType, $paramType](
-              ${paramName.toString.trim},
-              ${if (isValueClass) q"_.$paramName" else q"$idx"},
-              $repeated,
-              $magnoliaPkg.CallByNeed($ref),
-              $scalaPkg.Array($annList)
```
joroKr21, line 364: This was the bug: missing `..$annList`

propensive:
It seems like a lot of these little issues crept in... :(

PR #236, opened 2020-04-18
Add test for deriving instances with refined types if implicit is in scope

regadas:
I was debugging a possible issue involving refined types and though this would be a good test example.

### DIFF
--- a/tests/src/main/scala/tests.scala
+++ b/tests/src/main/scala/tests.scala
@@ -176,6 +176,14 @@ object Character {
   type Id = Long with Tag
 }
 
+case class AnotherCharacter(id: AnotherCharacter.Id)
+object AnotherCharacter {
+  trait Tag extends Any
+  type Id = Long with Tag
+
+  implicit val idShow: Show[String, Id] = _.toString
+}
+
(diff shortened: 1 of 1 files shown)


tests/src/main/scala/tests.scala
```diff
@@ -176,6 +176,14 @@ object Character {
   type Id = Long with Tag
 }
 
+case class AnotherCharacter(id: AnotherCharacter.Id)
+object AnotherCharacter {
+  trait Tag extends Any
+  type Id = Long with Tag
+
+  implicit val idShow: Show[String, Id] = _.toString
```
joroKr21, line 184: Hmm is this instance actually needed?

regadas: Hi @joroKr21 yeah it's needed otherwise we will hit this [case]( https://github.com/propensive/magnolia/blob/fa8b0d420f425fc97a37ee67ae8d2fc45e0d92c6/tests/src/main/scala/tests.scala#L653)

regadas:
@propensive np! Thanks for taking a look! 👍

Issue #61, opened 2017-12-06
Type classes with "exists" semantics


joroKr21:
What I mean is what if to create a type class for the whole case class / sealed trait I only need the instance for one field / subtype (or a subset of all fields / subtypes)? Like finding a field of some type deep within the structure.

E.g. `Default` only uses the first subtype instance. In fact it could be any instance. The rest of the subtype instances don't even need to exist.

I'm not sure if #46 would enable this?

propensive:
I think I understand, and I think *something* would be possible with #46, though if you have a concrete example, I would probably be more certain that I'm understanding correctly!

joroKr21:
One example is `Default` for a sealed trait:
```scala
sealed trait MaybeDate
case class ADate(date: Date) extends MaybeDate
case object NoDate extends MaybeDate
```
If I don't have a default `Date` I should still have a default `MaybeDate` by virtue of `NoDate`.

In general for any coproduct `A :+: B :+: C :+: CNil` there should be a `Default` instance whenever there exists a default instance for at least one of `A`, `B` or `C`. With shapeless I can express this like so (note the missing instance for `CNil`):
```scala
implicit def defaultLeft[L, R <: Coproduct](implicit L: Default[L]): Default[L :+: R] = ...
implicit def defaultRigth[L, R <: Coproduct](implicit R: Default[R]): Default[L :+: R] = ...
```

propensive:
That's a pretty clear example, thanks. I think we can offer a `dispatch` parameter which provides a non-empty list of derived typeclasses. (This will probably make more sense when I finally merge #46.) It needs a bit of reworking to ensure derivation continues even when not every branch of the coproduct can be derived, but should be doable.

PR #240, opened 2020-04-27
Allow derivation result to have arbitrary type

OlegYch:


joroKr21:
I am surprised that this was so easy.
I also wonder if this solves #61 🤔

OlegYch:
@joroKr21 it's just a fix for `dispatch`, `combine` was already supporting this
this will probably not work for full-auto derivation and recursive classes so doesn't fully address #107

OlegYch:
@joroKr21 i adjusted tests to better show what this enables

joroKr21:
Sure but I can imagine having mutual recursion.
And now we can also have e.g. `CsvColumn` and `CsvRow`.

propensive:
That's interesting. I remember that the type ascription has come and gone a few times, and I thought that we ended up keeping it because it was needed for deriving recursive types... but it was added or removed from both `dispatch` and `combine` in the past; not one of them. A recursive type will need _both_ `combine` and `dispatch`, so I guess that either one of these could have the ascription, and it's sufficient for the typechecker to terminate.

Thanks for the contribution, @OlegYch! I'll make a release now.

PR #248, opened 2020-05-08
Add check that all traits in a hierarchy are sealed

### LINKED ISSUE #243: Check in knownSubclasses that all subtraits are sealed
I'm wondering if we should support the following code:
```
  sealed trait Color7
  trait Color7a extends Color7
  object Color7 {
    case object Red extends Color7a
    case class Custom(rgb: String) extends Color7a
  }
```

By deriving `Color7`, magnolia throws a compilation error:
```
magnolia: could not find any direct subtypes of trait Color7
```
Is it a use-case we should support?

joroKr21:
... and that all classes are final or case classes.
Otherwise the act as an escape hatch (can be extended).'

And make use of `collect`.

Closes #243

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -100,6 +100,7 @@ object Magnolia {
     val repeatedParamClass = definitions.RepeatedParamClass
     val scalaSeqType = typeOf[Seq[_]].typeConstructor
     val javaAnnotationType = typeOf[java.lang.annotation.Annotation]
+    val typeClassName = TypeName("Typeclass")
 
     val prefixType = c.prefix.tree.tpe
     val prefixObject = prefixType.typeSymbol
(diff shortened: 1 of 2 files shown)


kubukoz:
Are classes also checked to be final?

joroKr21:
> Are classes also checked to be final?

No, but I think they should be checked for being either final or case classes.

joroKr21:
@propensive the build is green although tests are not passing.
Also the tests don't run on Scala 2.13 🤔

propensive:
I'll fix the problem with the tests soon, when I move to Github Actions.

propensive:
Also I approved (looks solid) though will wait until that last test actually passes before merging. ;)

joroKr21:
> Also I approved (looks solid) though will wait until that last test actually passes before merging. ;)

Done ✅

PR #260, opened 2020-06-10
Resolve alias for context typename (fixes #259)

### LINKED ISSUE #259: Typename.typeArguments are lost in case of type alias
When the type requesting a typeclass is somehow simplified using an alias, the expanded type parameters are not propagated to `TypeName` properly.

Steps to reproduce:
[Scastie](https://scastie.scala-lang.org/GeBK33EyQiqiHEpOii3noQ)
[Gist](https://gist.github.com/ivan-klass/18239911298ea112f5b853c63621752d)

ivan-klass:
See #259 
The test added by this PR is failing without the `.dealias` patch

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -305,7 +305,7 @@ object Magnolia {
         q"$magnoliaPkg.TypeName(${ts.owner.fullName}, ${ts.name.decodedName.toString}, $typeArgNames)"
       }
 
-      val typeNameDef = q"val $typeName = ${typeNameRec(genericType)}"
+      val typeNameDef = q"val $typeName = ${typeNameRec(genericType.dealias)}"
 
       val result = if (isRefinedType) {
         error(s"could not infer $prefixName.Typeclass for refined type $genericType")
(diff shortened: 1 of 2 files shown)


ivan-klass:
@propensive Sorry for bothering, could you please provide any feedback on this? We look forward for this fix as inner type shortcuts are important part of our DSL-like structures and dealiasing them manually is not an option. Thanks

ivan-klass:
Any updates on this? Still waiting for any feedback

joroKr21:
Hi @ivan-klass, sorry for the delay. This is a tricky one.

Sometimes dealiasing is what you expect, sometimes the opposite. Actually `scalac` has lots of issues open because it's not specified when dealiasing can happen and people get surprised.

Back to Magnolia land, I wonder if we should dealias `genericType` in the beginning rather than just for `typeName`.

propensive:
Very sorry for the delay, @ivan-klass. I didn't see the notification, unfortunately. I think we should merge this. I have vague memories of some of the dealiasing issues @joroKr21 mentioned, but my feeling is that dealiasing should produce _less surprising_ results in general. We can publish a new release, and roll back if it's problematic.

propensive (approved):

PR #264, opened 2020-08-15
A lot of refactoring

joroKr21:
Over time the code has become messy and duplicated.

 * Use object, type and symbol constants instead of untyped trees
 * Use constants from `definitions` where possible
 * Factor out `construct*` helper methods

### DIFF
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -81,31 +81,52 @@ object Magnolia {
     *  parameters apply equally to `dispatch` as to `combine`.
     *  */
   def gen[T: c.WeakTypeTag](c: whitebox.Context): c.Tree = Stack.withContext(c) { stack =>
-    import c.universe._
     import c.internal._
+    import c.universe._
+    import definitions._
+
+    val genericType = weakTypeOf[T]
+    val genericSymbol = genericType.typeSymbol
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/magnolia.scala
```diff
@@ -81,32 +81,53 @@ object Magnolia {
...
+    val ReadOnlyParamObj = reify(ReadOnlyParam).tree
+    val ReadOnlyParamSym = symbolOf[ReadOnlyParam[Any, Any]]
+    val RightObj = reify(Right).tree
+    val SealedTraitSym = symbolOf[SealedTrait[Any, Any]]
+    val SeqTpe = typeOf[Seq[Any]].typeConstructor
+    val SomeObj = reify(Some).tree
+    val SubtypeObj = reify(Subtype).tree
+    val SubtypeTpe = typeOf[Subtype[Any, Any]].typeConstructor
+    val TypeClassNme = TypeName("Typeclass")
+    val TypeNameObj = reify(magnolia.TypeName).tree
```
propensive, line 117: This looks much cleaner. Do you think there's any performance benefit to making them lazy?

joroKr21: I was wondering the same. The cost is wrapping them in `LazyRef`. Do you think we should?

propensive: Let's make it a future enhancement. ;)

PR #270, opened 2020-09-27
Exclude params and cons from the derivation based on annotations

vigoo:
This is a feature request with possible proof-of-concept implementation. I would be happy to work on an alternative approach if there is any.

### The problem
The real-world example for the requested feature came up while trying to migrate the semi-auto derived _binary codecs_ in https://github.com/vigoo/desert/ from shapeless to magnolia. It supports marking case class fields and sealed trait constructors as _transient_. These transient subtrees of a data type may have types for which there is no implicit binary codec available and it cannot be automatically derived either. Even though the transient annotations can be observed from magnolia's `combine`/`dispatch` methods, as the derivation engine does not know about these application-specific exclusion rules it tries to gather typeclass instances for these members too, and fails.

### A possible solution
In this PR I created a possible solution for the above problem, while trying to keep it as generic as possible.
What I came up with is the following:

- There is a marker trait `RuntimeBoundTypeclass` to be mixed in to `StaticAnnotations` 
- This is used to communicate with the derivation engine that we don't need to look for the typeclass instance for a given parameter or subclass
- In order to not break the API we still need to have a `Typeclass[T]` value somehow. What I did is to defer getting the value to a user-defined function `bind[T]: Typeclass[T]` in the enclosing object.

For the use case I described above, `bind` will never be called so it can just throw an exception; `combine` and `dispatch` can detect the annotations and simply not call `.typeclass` that is bound to `bind`.
I also included a simpler example that is based on the `Default` typeclass's example.

Note that this provides an alternative solution for https://github.com/propensive/magnolia/issues/61 too

### DIFF
--- a/core/shared/src/main/scala/interface.scala
+++ b/core/shared/src/main/scala/interface.scala
@@ -15,7 +15,7 @@
 package magnolia
 
 import language.experimental.macros
-import scala.annotation.tailrec
+import scala.annotation.{StaticAnnotation, tailrec}
 import mercator._
 
 /** represents a subtype of a sealed trait
@@ -455,8 +455,7 @@ object MagnoliaUtil {
 
 /**
  * Marker trait for annotations which when encountered on either a parameter or a constructor,
- * the derivation will not look for a typeclass instance, instead bind the typeclass parameter
- * on the interface to the 'bind[A]: Typeclass[A]' method of the enclosing class.
- *
+ * the derivation will not look for a typeclass instance but use the 'fallback[T]: Typeclass[T]'
+ * method of the enclosing class instead.
  * */
-trait RuntimeBoundTypeclass
+trait ForceFallbackDerivation extends StaticAnnotation
--- a/core/shared/src/main/scala/interface.scala
+++ b/core/shared/src/main/scala/interface.scala
@@ -452,3 +452,11 @@ object MagnoliaUtil {
     values.toList.collect { case Left(v) => v }
 
 }
+
+/**
+ * Marker trait for annotations which when encountered on either a parameter or a constructor,
+ * the derivation will not look for a typeclass instance, instead bind the typeclass parameter
+ * on the interface to the 'bind[A]: Typeclass[A]' method of the enclosing class.
+ *
+ * */
+trait RuntimeBoundTypeclass
--- a/core/shared/src/main/scala/interface.scala
+++ b/core/shared/src/main/scala/interface.scala
@@ -458,4 +458,4 @@ object MagnoliaUtil {
  * the derivation will not look for a typeclass instance but use the 'fallback[T]: Typeclass[T]'
  * method of the enclosing class instead.
  * */
-trait ForceFallbackDerivation extends StaticAnnotation
+trait ExcludeFromDerivation extends StaticAnnotation
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -261,12 +261,12 @@ object Magnolia {
       case _ => false
     }
 
-    def isRuntimeBoundTypeclassAnnotation(typ: Tree): Boolean = {
-      val e = c.Expr[Any](c.typecheck(q"new magnolia.TypeCheckHelper[$typ]"))
+    def isForceFallbackDerivationAnnotation(typ: Tree): Boolean = {
+      val e = c.Expr[Any](c.typecheck(q"new _root_.magnolia.TypeCheckHelper[$typ]"))
       val t = e.actualType match {
         case TypeRef(_, _, params) => params.head
       }
-      t <:< typeOf[RuntimeBoundTypeclass]
+      t <:< typeOf[ForceFallbackDerivation]
     }
 
     val expandDeferred = new Transformer {
@@ -450,13 +450,13 @@ object Magnolia {
                   typeclassTree(paramType, typeConstructor, ref)
                 }
 
-                val isRuntimeBound = anns.exists {
-                  case Apply(Select(New(typeTree), _), _) if isRuntimeBoundTypeclassAnnotation(typeTree) => true
+                val isForcedFallback = anns.exists {
+                  case Apply(Select(New(typeTree), _), _) if isForceFallbackDerivationAnnotation(typeTree) => true
                   case _ => false
                 }
                 val assigned =
-                  if (isRuntimeBound) {
-                    q"def $ref = ${c.prefix}.bind[$paramType]"
+                  if (isForcedFallback) {
+                    q"def $ref = ${c.prefix}.fallback[$paramType]"
                   } else {
                     deferredVal(ref, searchType, derivedImplicit.fold(error, identity))
                   }
@@ -598,13 +598,13 @@ object Magnolia {
           val anns = annotationsOf(subType.typeSymbol)
           val path = CoproductType(genericType.toString)
           val frame = stack.Frame(path, resultType, assignedName)
-          val isRuntimeBound = anns.exists {
-            case Apply(Select(New(typeTree), _), _) if isRuntimeBoundTypeclassAnnotation(typeTree) => true
+          val isForcedFallback = anns.exists {
+            case Apply(Select(New(typeTree), _), _) if isForceFallbackDerivationAnnotation(typeTree) => true
             case _ => false
           }
 
-          subType -> (if (isRuntimeBound) {
-            q"${c.prefix}.bind[$subType]"
+          subType -> (if (isForcedFallback) {
+            q"${c.prefix}.fallback[$subType]"
           } else {
             stack.recurse(frame, appliedType(typeConstructor, subType), shouldCache) {
               typeclassTree(subType, typeConstructor, termNames.ERROR)
--- a/core/shared/src/main/scala/magnolia.scala
+++ b/core/shared/src/main/scala/magnolia.scala
@@ -261,12 +261,12 @@ object Magnolia {
       case _ => false
     }
 
-    def isForceFallbackDerivationAnnotation(typ: Tree): Boolean = {
+    def isExcludeFromDerivationAnnotation(typ: Tree): Boolean = {
       val e = c.Expr[Any](c.typecheck(q"new _root_.magnolia.TypeCheckHelper[$typ]"))
       val t = e.actualType match {
         case TypeRef(_, _, params) => params.head
       }
-      t <:< typeOf[ForceFallbackDerivation]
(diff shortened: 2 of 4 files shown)


core/shared/src/main/scala/interface.scala
```diff
@@ -452,3 +452,11 @@ object MagnoliaUtil {
...
 
 }
+
+/**
+ * Marker trait for annotations which when encountered on either a parameter or a constructor,
+ * the derivation will not look for a typeclass instance, instead bind the typeclass parameter
+ * on the interface to the 'bind[A]: Typeclass[A]' method of the enclosing class.
+ *
+ * */
+trait RuntimeBoundTypeclass
```
propensive, line 462: Should this be self-typed to `StaticAnnotation`?, i.e.
```scala
trait RuntimeBoundTypeclass { this: StaticAnnotation =>
```
(I'm basing that only on what you said in the PR description!)

core/shared/src/main/scala/magnolia.scala
```diff
@@ -261,6 +261,14 @@ object Magnolia {
       case _ => false
     }
 
+    def isRuntimeBoundTypeclassAnnotation(typ: Tree): Boolean = {
+      val e = c.Expr[Any](c.typecheck(q"new magnolia.TypeCheckHelper[$typ]"))
```
propensive, line 265: This should be `_root_.magnolia.TypeCheckHelper`

core/shared/src/main/scala/magnolia.scala
```diff
@@ -577,11 +595,21 @@ object Magnolia {
...
-          subType -> stack.recurse(frame, appliedType(typeConstructor, subType), shouldCache) {
-            typeclassTree(subType, typeConstructor, termNames.ERROR)
-          }.fold(error, identity)
+          val isRuntimeBound = anns.exists {
+            case Apply(Select(New(typeTree), _), _) if isRuntimeBoundTypeclassAnnotation(typeTree) => true
+            case _ => false
+          }
+
+          subType -> (if (isRuntimeBound) {
+            q"${c.prefix}.bind[$subType]"
```
propensive, line 607: Can we use `fallback` instead of `bind`? I didn't ever use the `fallback` method, but I think it serves the same purpose.

core/shared/src/main/scala/interface.scala
propensive: I might also suggest a better name for this. If we use the `fallback` method instead of `bind`, could we call it `FallbackDerivation` or just `Fallback`?

propensive (changes requested):
I think it looks reasonable! @joroKr21 may have some thoughts, too.

Could you convert the `println`s into unit tests so they get run when the other unit tests are run?

propensive:
I've just changed Magnolia's default branch from `master` to `main`, which is not just a name-change, but actually changes a lot of the structure of the repository. If you want to make changes to this PR, please continue without rebasing or merging. I will handle the merge/rebase when it's ready.

vigoo:
Ok thank you! I'll get back to this soon

core/shared/src/main/scala/magnolia.scala
vigoo: I did not notice before that there is `fallback`. I think it's a good idea to reuse it for this use case too, then instead of talking about "runtime bound" it could be "forced fallback".

vigoo:
@propensive I changed the PR to reuse the existing `fallback` method and changed the naming accordingly, and added the example to the tests.

joroKr21:
I've been thinking about this a bit - I'm not sure reusing `fallback` is a good idea. The problem - we are conflating semantics here. `fallback` will be used when there is no instance available (even when there is no annotation). So if you just want to ignore certain fields you will also get the fallback whether you wanted it or not 😕

joroKr21:
Why can't we just create a `CallByNeed` that fails - I don't think we need another special method.
About naming - I think something along the lines of `@exclude`, `@ignore` or `@noDerive` make sense.
And it would make it more usable if we have a corresponding boolean method `isExcluded` / `isIgnored` etc.

joroKr21:
Unfortunately it doesn't quite solve #61 because it would have to be more "dynamic" rather than "static" in a sense that it should be based on the available implicits in scope.

vigoo:
Having a built-in `@exclude` or `@ignore` or `@noDerive` would not solve my initial problem, this should be controlled by use-case specific annotations. For example in the serialization library I would like to use this, `@TransientField(default)` has a default value parameter - but I believe this should remain specific to that library and not in magnolia. That's why I came up with the marker trait for annotations. 

Then tried to figure out what else it could be used for, to make sure I'm adding something generally useful. That's how I ended with the notion of "runtime binding" type classes in places marked by annotations having the given trait. This is more generic than my use case, but enables it by creating a bind that does nothing. I did not notice the fallback mechanism at the time. Then to avoid having two very similar things I changed this to depend on the fallback function. I also have mixed feelings about this as its a bit of mix of semantics as you wrote. To avoid the fallback getting called for missing things, my original version with the new callback function may be better.

I guess if I don't try to be this generic here and only solve my particular problem then instead of the callback it could be just an implicit failing binder and then the 'marker trait' part would remain but there would be no extra method to be added to the user code. Is this what you suggest?

joroKr21:
`@TransientField(default)` - can't you just use default parameters with an `@exclude` annotation?
I'm worried that API is becoming more and more complicated.

> there would be no extra method to be added to the user code

Yes, definitely that. Take a look at `CallByNeed`. You can just create one that fails - the user would be responsible for checking the "marker" annotation (or better yet, we add a method that does that) and not trying to get the instance.

vigoo:
@joroKr21 I believe I changed it in the way you recommended.

propensive:
> And it would make it more usable if we have a corresponding boolean method `isExcluded` / `isIgnored` etc.

I'm not sure it's worth it. There's a small cost in derivation time for every additional method we add to the `Param` interface...

propensive:
Would it be possible to use a marker trait on the fields which should be ignored?
```
trait Ignored
```
and provide a higher-priority instance in the derivation object which you can do a reference equality check against inside `combine` to check whether the instance-corresponding-to-`Ignored` was chosen, and then ignore those fields?

```
implicit val ignored: Tc[_ <: Ignored] = <something>
```
then inside `combine`,
```
if(param.typeclass eq ignored) // skip it
```

The biggest problem is probably going to be that you have `with Ignored` littering all your parameters, though doesn't shapeless offer some sort of "type tagging" that could work around that problem?

propensive:
As a potentially more powerful solution, could we use a `@typeclasses(_*)` annotation which provides a sequence of typeclass instances that could be specified on a case class field, and tried as possible solutions *before* doing implicit search?

Here's the definition (to be included in Magnolia):
```
case class typeclasses[T](xs: T*) extends StaticAnnotation
```

And you'd use it like this,
```
case class MyProduct(x: Int, y: String, @typeclasses(ignoredShow, ignoredDecoder) z: Int)
```

We would just need to modify Magnolia to check if the annotation exists on that field and contains an instance that conforms to the search type, then use that instead of doing implicit search.

In your particular derivation, you'd annotate with `@typeclasses(someIgnoredInstance)` and do the equality check I suggested above.

vigoo:
@propensive I think your ideas are possible workarounds but does not really satisfy the goal I tried to solve. Naturally, I respect if you think this does not fit well into magnolia and I'll stop trying to push this :) But I feel that maybe I did not communicate well enough what I want to achieve so I'll try to give some more context in this comment.

I have a binary serialization library (linked above) with the main focus on supporting evolution of data types over time, and it is used in production for serializing Akka remoting messages and Akka persistence events/snapshots. The idea is that it's a "semi-auto" kind of derivation where you list the history of your data types by listing changes for the derive function such as "field added", "field removed", etc. Transient fields and constructors are necessary for this use case; the easiest way to model transient state in persistent typed akka actors is to have transient fields in the persistent state. And it helps a lot to have transient constructors in actor message types for messages that are guaranteed to be sent only locally. In both of these cases you may use types that you cannot or don't want to define a binary codec for. In my library this is currently implemented with shapeless and works well; I wanted to migrate to magnolia because the shapeless based code is very complex and hard to maintain, and because with a higher level derivation abstraction I had the feeling that magnolia would have a Scala 3 release pretty soon and then I get "free" Scala 3 support without reimplementing the derivation with Scala 3's low level derivation support. This idea was blocked by not being able to handle the transient fields and constructors in any way (at least I could not find any) so I tried to find a way to add it to magnolia with making it useful to other use cases as much as possible. 

All the ideas listed above:
- using default values instead of ones defined in the annotation
- tagging the type itself 

are breaking the client code that I no longer maintain and I would not want to do such breaking changes. For my particular use case it's better to just maintain a separate scala 3 version of the generic binary codec instead.

That said I would not have open this PR if I did not believe this is something useful for other use cases; 

About the last idea of the `@typeclasses` annotation, I think it's a good idea, but in my opinion it leaks implementation details to the user code. If magnolia is used by libraryX to provide derivation of a type class defined in libraryX, the user of that library should not be aware of magnolia-specific annotations or concepts; If "cutting" the derivation tree is part of the library's logic then in my opinion it should be controllable from the derivation implementation, so for the end users of it is expressed in the domain of the library.

propensive:
I have some of the same concerns about scope creep of the API, like @joroKr21, but I see that as generally less important than your use case (which I think is perfectly valid) so don't worry about that! I know I've also only been paying cursory attention to this (and haven't paid much attention to the code) so apologies if I was kind of ignoring the point and shooting random ideas!

In any case, I agree with your point about the `@typeclasses` suggestion... It would have been more powerful in some ways, but it's still a bit hacky. I just wanted to pitch for the most expressive approach we could achieve.

So, going back to the actual implementation, my (new, seeing as I suggested it originally) concern about the reuse of `fallback` is that derivation will succeed for types where it shouldn't: non-ignored fields of types which have no typeclass instance, even though it would work fine for your use case for products that really should be derived.

I also think the annotation mixin is an interesting way to do this, though a little inventive. So having a default `@skipDerivation` which subtypes `RuntimeBoundTypeclass` and `StaticAnnotation` (or something with a similar name) would be good. I'm open to suggestions on the name.

So I'm more or less happy with the approach. I'd like to take some time to give the code a proper review, but I think we're basically good.

vigoo:
Than you! You will see that the latest version of the PR is less powerful than the initial suggestion by only supporting exclusion and not arbitrary dynamic bindings, but this way it does not need any new user-defined method or the use of fallback.

joroKr21:
https://github.com/propensive/magnolia/pull/279#issuecomment-766475143

PR #274, opened 2020-11-10
Add `@performance` annotation

propensive:
This introduces a new `@performance` annotation, which can be used on a `gen` method definition (the same as a `@debug` annotation) to trigger a performance report (provided by Probably) to be displayed once at the end of compilation. A single `@performance` annotation will trigger performance reporting for _all_ derivations (by design).

This PR relies on a trick suggested by `@som-snytt`.

Still to do: Make Probably an optional dependency of Magnolia. It should not be required in all cases.


propensive:
An example of a performance report can be seen in GitHub Actions.

joroKr21:
No clue how to read fury output but this looks great!

PR #275, opened 2020-12-06
Draft for Scala 3

maphi:
**To build this use sbt, not fury!**

Current state:

Missing/Unresolved:
- subtyping for sealed traits (wasn't able to get that to work yet, but should be doable)
- uses sbt instead of fury (I lack experience with fury, help needed)
- construct methods for product types missing
- for some annotations it is not clear to me where they would even be written (see TODO in PR)
- derive opaque types (not clear if this would be possible at all)
- derive value classes (not clear if this would be possible at all)
- get TypeInfo may not be exhaustive for all possible types yet (see code)
- scala 2 compatibility, adjust scala 2 code
- mercator for scala 3, adjust stuff outside off core

Changes:
- derivation is done via trait so that methods combine/dispatch are easily accessible
- rename gen to derived to match Scala 3 derivation
- Typename was replaced by TypeInfo


propensive:
This is wonderful to see, thank you! And it's come at a good time: I have good availability this week to experiment with this myself, so I should be able to get into the details and attempt some of the missing parts.

propensive:
Superceded by #281.

Issue #266, opened 2020-09-03
Backward incompatibility of 0.17.0


catostrophe:
#257 has broken the backward compatibility of `magnolia`.

When I bumped version, I got a runtime error in a lib dependent on 0.16.0:
```
java.lang.NoSuchMethodError: 'magnolia.Param magnolia.Param$.apply(java.lang.String, int, boolean, magnolia.CallByNeed, magnolia.CallByNeed, java.lang.Object[])'
```

The newly added param should have been with the default value.

joroKr21:
To promise backwards compatibility we would have to release 1.0

catostrophe:
Magnolia has become a widely used library. The thing is that in a relatively small project it comes as a transitive dependency from 4 different libs.

If only one lib updates magnolia to 0.17.0, and the others don't come up with new releases, the whole app may be spoilt silently (will break at runtime). That's what really happened.

It doesn't cost much to support backward compatibility when you add a new argument such as an array of annotations.

joroKr21:
I'm not even against releasing 1.0 but if we go through with #247 and #256 they will for sure not be binary compatible.
And the only way to be sure is to add MiMa to the build.

propensive:
Thanks for reporting, @catostrophe. I want to move to version 1.0 quite soon now, though it does need a bit of admin work, which I haven't had time for.

propensive:
#247 _can_ actually be binary compatible, as the methods aren't part of any binary API. But in any case, I would support `split` and `join` as aliases for `dispatch` and `combine`.

PR #279, opened 2021-01-24
Add possibility to separate `gen` from `combine`

oyvindberg:
This solves two related issues with magnolia:

### auto/semiauto

Some people prefer automatic derivation, while others prefer semiauto. As it stands now the type class author has to decide which one to offer (based on the existence of `implicit` of the `gen` method which is implemented by the magnolia macro)

This choice should be made by users instead. The mechanism introduced in this PR enables a type class author to offer both

### Bundling more than one type class derivation behind one import

As it stands now, it's not really possible to offer more than one type class derivation through a wildcard import, because they ultimately need to be implemented in an `object`, where the `combine` methods and so on would clash.

## Workarounds 

If you look around you'll find that many projects have workarounds by writing their own macros around the magnolia macro. For instance:
- https://github.com/spotify/magnolify/blob/master/cats/src/main/scala/magnolify/cats/auto/CatsMacros.scala#L30
- https://github.com/softwaremill/tapir/blob/master/core/src/main/scala/sttp/tapir/generic/auto/package.scala#L9

## Proposed solution

Add an annotation to `gen` where we can specify the `object` which has the `combine` method and the rest of the machinery.
For now this annotation is called `@proxy`, but I'm sure there are better names. 

Example, given two non-interesting type classes called `ToString1` and `ToString2`:

```scala

  object auto {
    @proxy(ToString1)
    implicit def deriveToString1[A]: ToString1[A] = macro Magnolia.gen[A]

    @proxy(ToString2)
    implicit def deriveToString2[A]: ToString2[A] = macro Magnolia.gen[A]
  }

  object semiauto {
    @proxy(ToString1)
    def deriveToString1[A]: ToString1[A] = macro Magnolia.gen[A]

    @proxy(ToString2)
    def deriveToString2[A]: ToString2[A] = macro Magnolia.gen[A]
  }

```

### DIFF
--- a/src/core/interface.scala
+++ b/src/core/interface.scala
@@ -440,6 +440,13 @@ final case class TypeName(owner: String, short: String, typeArguments: Seq[TypeN
   */
 final class debug(typeNamePart: String = "") extends scala.annotation.StaticAnnotation
 
+/**
+  * Allows `combine`, `dispatch` and `fallback` to be implemented in another object.
+  * This makes it possible to export more than one derivation in an object
+  * @param companion the object where the methods are defined
+  */
+final class proxy(companion: Any) extends scala.annotation.StaticAnnotation
+
 private[magnolia] final case class EarlyExit[E](e: E) extends Exception with util.control.NoStackTrace
 
 object MagnoliaUtil {
--- a/src/core/magnolia.scala
+++ b/src/core/magnolia.scala
@@ -106,6 +106,7 @@ object Magnolia {
     val NoneObj = reify(None).tree
     val ParamObj = reify(Param).tree
     val ParamSym = symbolOf[Param[Any, Any]]
+    val ProxyTpe = typeOf[proxy]
     val ReadOnlyCaseClassSym = symbolOf[ReadOnlyCaseClass[Any, Any]]
     val ReadOnlyParamObj = reify(ReadOnlyParam).tree
     val ReadOnlyParamSym = symbolOf[ReadOnlyParam[Any, Any]]
@@ -118,7 +119,13 @@ object Magnolia {
     val TypeClassNme = TypeName("Typeclass")
     val TypeNameObj = reify(magnolia.TypeName).tree
 
-    val prefixType = c.prefix.tree.tpe
+    val proxy: Option[c.Expr[_]] =
+      c.macroApplication.symbol.annotations.collectFirst {
+        case a if a.tree.tpe <:< ProxyTpe => c.Expr(a.tree.children(1))
+      }
+
+    val prefix: Expr[_] = proxy.getOrElse(c.prefix)
+    val prefixType = prefix.tree.tpe
     val prefixObject = prefixType.typeSymbol
     val prefixName = prefixObject.name.decodedName
 
@@ -234,7 +241,7 @@ object Magnolia {
 
     def extractParameterBlockFor(termName: String, category: String): List[Symbol] = {
       val term = TermName(termName)
-      val classWithTerm = c.prefix.tree.tpe.baseClasses
+      val classWithTerm = prefix.tree.tpe.baseClasses
         .find(cls => cls.asType.toType.decl(term) != NoSymbol)
         .getOrElse(error(s"the method `$termName` must be defined on the derivation $prefixObject to derive typeclasses for $category"))
 
@@ -380,7 +387,7 @@ object Magnolia {
 
         val impl = q"""
           $typeNameDef
-          ${c.prefix}.combine(new $caseClassType(
+          $prefix.combine(new $caseClassType(
             $typeName,
             true,
             false,
@@ -549,7 +556,7 @@ object Magnolia {
             val $paramsVal = new $ArrayClass[$paramType](${assignments.length})
             ..$assignments
             $typeNameDef
-            ${c.prefix}.combine(new $caseClassType(
+            $prefix.combine(new $caseClassType(
               $typeName,
               false,
               $isValueClass,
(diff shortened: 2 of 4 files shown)


joroKr21:
We have a lot of requests to customise the derivation but keep reaching for annotations when we don't know what to do.
I suggest a way to configure the macro:

```scala
  trait Config {
    type Proxy <: Singleton { type Typeclass[A] }
    type Ignore <: annotation.Annotation
    val readOnly: Boolean
    val minFields: Int
    val maxFields: Int
    val minCases: Int
    val maxCases: Int
  }

  def genWith[T: c.WeakTypeTag, C <: Config with Singleton: c.WeakTypeTag](c: whitebox.Context): c.Tree = {
    import c.internal._
    import c.universe._

    weakTypeOf[C].decls.toList.map(x => x -> x.typeSignature).foreach(println)
    Magnolia.gen[T](c)
  }

  object CsvConfig extends Magnolia.Config {
    type Proxy = Csv.type
    type Ignore = transient
    final val readOnly = true
    final val minFields = 0
    final val maxFields = -1
    final val minCases = 0
    final val maxCases = -1
  }

  implicit def deriveCsv[A]: Csv[A] = macro Magnolia.genWith[A, CsvConfig.type]

/*
(constructor CsvConfig,()magnolia.examples.Csv.CsvConfig.type)
(type Proxy,magnolia.examples.Csv.type)
(type Ignore,transient)
(value readOnly,=> Boolean(true))
(value readOnly,Boolean(true))
(value minFields,=> Int(0))
(value minFields,Int(0))
(value maxFields,=> Int(-1))
(value maxFields,Int(-1))
(value minCases,=> Int(0))
(value minCases,Int(0))
(value maxCases,=> Int(-1))
(value maxCases,Int(-1))
*/
```

oyvindberg:
Hey there @joroKr21 - sorry about the delay.

I'll preface by saying I would accept any encoding for this, including the one you proposed.

That said, I don't particularly think annotations are a bad fit for this. Most people have negative associations with annotations because how people use them at runtime - but as a compile-time thing I think they are completely fine! Especially for this particular usage, to pass static information to a macro. They are easy to use and easy to document.

I'm a bit skeptical towards the `Config` because you are immediately exposed to all options, even if they don't apply. The options themselves, however, are only valid by themselves in more narrow contexts. For instance:

- `@debug` and `@proxy` apply to the macro expansion itself
- `@maxFields` and `@minFields` belong with the implementation of `combine` (if any), and should absolutely not be configurable at macro-invocation time.
- `@maxCases` and `@minCases` I imagine would belong with the implementation of `dispatch` (if any).

I don't exactly know where `@ignore` would fit in, probably along with `@debug`/`@proxy` as something supplied by the user?

joroKr21:
I may be traumatised by Spring when it comes to configuration with annotations 😃 

> I'm a bit skeptical towards the Config because you are immediately exposed to all options, even if they don't apply.

I actually like that because you can see all configuration options at a glance and you have autocomplete etc.
Note that we can provide default configuration so you wouldn't necessarily have to specify all options.

> The options themselves, however, are only valid by themselves in more narrow contexts.

That's a valid point. Still I'm not sure which is better or worse.
Another thing we could go for is a `@configure` annotation with all options.

oyvindberg:
> I may be traumatised by Spring when it comes to configuration with annotations smiley

Aren't we all :)



The `Config` trait actually has two further problems I think:

- Backwards/forwards compatibility if we add more parameters, adding a member to a trait is not backwards compatible.
- Your code seems to rely on lifting the values to the type level (via `final`), right? At that point you really want `type maxCases <: Int`, or if you account for optionality `type maxCases <: Option[Int] = None` with the override `override type maxClasses = Some[1]`? 

If it's the discoverability we want we could scope all the annotations inside a `object`maybe?

>  Another thing we could go for is a @configure annotation with all options.

I like this better, but there will still be backwards compatibility issues when adding a parameter, unless we provide an overloaded constructor.

plokhotnyuk:
@oyvindberg Am I understand right that if `Config` is used only in compile time then it is more about the _source_  compatibility (no binary compatibility issues through transitive dependencies)? Another question about compile time annotations: will they be supported by Scala 3?

oyvindberg:
> Am I understand right that if Config is used only in compile time then it is more about the source compatibility (no binary compatibility issues through transitive dependencies)?

@plokhotnyuk As long as it doesn't cross any sbt project boundary (where an old version of magnolia might have been evicted in favour of a newever version in a downstream project) I suppose that's true. I'm just wary I suppose after seeing #266 .


> Another question about compile time annotations: will they be supported by Scala 3?

I would be incredibly surprised if they didn't, but let's find out!

plokhotnyuk:
In [jsoniter-scala](https://github.com/plokhotnyuk/jsoniter-scala) for configuration of codec derivation I've used [a class with a private constructor and builder-like methods](https://github.com/plokhotnyuk/jsoniter-scala/blob/4aedead8da0b6ab7178f74cd685815150fd08721/jsoniter-scala-macros/shared/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/macros/JsonCodecMaker.scala#L26-L206). The [expression of configuration is passed to the `make` macro](https://github.com/plokhotnyuk/jsoniter-scala/blob/4aedead8da0b6ab7178f74cd685815150fd08721/jsoniter-scala-macros/shared/src/test/scala/com/github/plokhotnyuk/jsoniter_scala/macros/JsonCodecMakerSpec.scala#L268-L269) where it is [instantiated using `Eval.eval`](https://github.com/plokhotnyuk/jsoniter-scala/blob/4aedead8da0b6ab7178f74cd685815150fd08721/jsoniter-scala-macros/shared/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/macros/JsonCodecMaker.scala#L371) which is [not supported in Scala 3](https://github.com/scalacenter/scala-3-migration-guide/pull/11).

oyvindberg:
So I did some hacking on a scala 3 magnolia here one afternoon, and so far my impression is that compile-time annotations work just fine, and the `@proxy` pattern suggested in this PR seems to work.

In fact, if we want to cross-compile code which uses magnolia between Scala 2 and 3 I think we'll *have* to do something in that direction, to separate `combine`/`dispatch` from `gen`, since we don't want the `macro` keyword to appear in Scala 3 code.

PR #281, opened 2021-04-08
Support for Scala 3

propensive:
This PR introduces support for Scala 3 to Magnolia, and builds on work by @kubukoz, @KacperFKorban, @maphi and others. It is essentially a complete rewrite, supported by VirtusLab.

The API has changed slightly from Magnolia for Scala 2 in the following ways:
- the `Typeclass` type is no longer required, but,
- derivation objects should extend `Derivation[Typeclass]` where `Typeclass` is the type constructor, formerly defined as a type member
- the `combine` method has been renamed to `join`, and its signature must conform to the `Derivation` interface
- the `dispatch` method has been renamed to `split`, and its signature must conform to the `Derivation` interface
- a default implementation of `split` is provided, so any user-specified `split` implementation must be annotated with `override`
- the `parameters` method is now called `params`
- the method which invokes the Magnolia macro no longer needs to be specified

Default parameters and value classes are not yet fully supported.


fsvehla:
@propensive Is accessing the default value of case classes a limitation of Scala 3? Is there a bug that we can track?

propensive:
I'm not sure, actually... I'll check with @KacperFKorban, and we should definitely track it in an issue if it can't be implemented right away.

heksesang:
@fsvehla The macro API in Scala 3 lacks a way to access the default value, but someone were working on adding it – https://github.com/lampepfl/dotty/pull/11667.

PR #283, opened 2021-04-20
Add TypeName to Param

stacycurl:
JIT created issue: https://github.com/propensive/magnolia/issues/282


propensive (approved):
It looks like a useful addition to me, and quite a straightforward change. I'm going to assume that performance won't be heavily affected, but I'm not sure we have a good way to test that any more...

stacycurl:
The build timed out, hope that wasn't due to performance being affected.

PR #291, opened 2021-05-17
Update of readme to working example

wojciechUrbanski:
The previous example in README did not work. I have aligned the naming convention with the `examples` (Show -> Print) and changed a little bit code so it compiles and works.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -19,18 +19,15 @@ __Magnolia__ is a generic macro for automatic materialization of typeclasses for
 
 Given an ADT such as,
 ```scala
-enum Tree[+T]:
+enum Tree[+T] derives Print:
   case Branch(left: Tree[T], right: Tree[T])
   case Leaf(value: T)
-
-object Tree:
-  given [T: [X] =>> Print[X]] : Print[Tree[T]] = Print.derived
 ```
 and provided an given instance of `Print[Int]` is in scope, and a Magnolia derivation for the `Print` typeclass
 has been provided, we can automatically derive given typeclass instances of `Print[Tree[Int]]` on-demand, like
 so,
 ```scala
-Print.derived[Tree[Int]].print(Branch(Branch(Leaf(1), Leaf(2)), Leaf(3)))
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -20,38 +20,44 @@ __Magnolia__ is a generic macro for automatic materialization of typeclasses for
 Given an ADT such as,
 ```scala
 enum Tree[+T]:
```
KacperFKorban, line 22: Wouldn't adding `derives` clause to the enum be more 'user friendly' (than an explicit given)?
```suggestion
enum Tree[+T] derives Print:
  case Branch(left: Tree[T], right: Tree[T])
  case Leaf(value: T)
```

readme.md
```diff
@@ -20,38 +20,44 @@ __Magnolia__ is a generic macro for automatic materialization of typeclasses for
...
+  given [T: [X] =>> Print[X]] : Print[Tree[T]] = Print.derived
 ```
-and provided an given instance of `Show[Int]` is in scope, and a Magnolia derivation for the `Show` typeclass
-has been provided, we can automatically derive given typeclass instances of `Show[Tree[Int]]` on-demand, like
+and provided an given instance of `Print[Int]` is in scope, and a Magnolia derivation for the `Print` typeclass
+has been provided, we can automatically derive given typeclass instances of `Print[Tree[Int]]` on-demand, like
 so,
 ```scala
-Branch(Branch(Leaf(1), Leaf(2)), Leaf(3)).show
+Print.derived[Tree[Int]].print(Branch(Branch(Leaf(1), Leaf(2)), Leaf(3)))
```
adamw, line 33: this does not compile after copy-paste - either an import (`Tree.*`) or qualifying the names e.g. `Tree.Branch` is necessary

wojciechUrbanski: I forgot to put it there... :) Fixed in new commit

readme.md
wojciechUrbanski: Definitely! :) I have changed it according to Your suggestion :)

PR #295, opened 2021-05-18
Fix owner name

### LINKED ISSUE #293: TypeInfo macro returns incorrect owner with Scala 3
Just sharing with you something I discovered while working on Scala 3 support for Caliban (I reused some code from the magnolia macros): the owner returned by `typeInfo` is incomplete. It just returns one level.

I ended up making it recursive like this:
```scala
  def typeInfoImpl[T: Type](using qctx: Quotes): Expr[TypeInfo] = {
    import qctx.reflect._

    def normalizedName(s: Symbol): String = if s.flags.is(Flags.Module) then s.name.stripSuffix("$") else s.name
    def name(tpe: TypeRepr) : Expr[String] = Expr(normalizedName(tpe.typeSymbol))

    def owner(tpe: TypeRepr): Expr[String] = {
      def loop(s: Symbol): String =
        if s.maybeOwner.isNoSymbol then ""
        else if (s.owner == defn.EmptyPackageClass) ""
        else if (s.owner == defn.RootClass) ""
        else {
          val parent = loop(s.owner)
          val self = normalizedName(s.owner)
          if(parent.isEmpty) self else s"$parent.$self"
        }
      Expr(loop(tpe.typeSymbol))
    }

    def typeInfo(tpe: TypeRepr): Expr[TypeInfo] = tpe match
      case AppliedType(tpe, args) =>
        '{TypeInfo(${owner(tpe)}, ${name(tpe)}, ${Expr.ofList(args.map(typeInfo))})}
      case _ =>
        '{TypeInfo(${owner(tpe)}, ${name(tpe)}, Nil)}

    typeInfo(TypeRepr.of[T])
  }
```
I know very little about macros so this might break in some cases, but hope it helps!

KacperFKorban:
fixes #293

### DIFF
--- a/src/core/macro.scala
+++ b/src/core/macro.scala
@@ -102,10 +102,14 @@ object Macro:
     def normalizedName(s: Symbol): String = if s.flags.is(Flags.Module) then s.name.stripSuffix("$") else s.name
     def name(tpe: TypeRepr) : Expr[String] = Expr(normalizedName(tpe.typeSymbol))
 
-    def owner(tpe: TypeRepr): Expr[String] =
-      if tpe.typeSymbol.maybeOwner.isNoSymbol then Expr("<no owner>")
-      else if (tpe.typeSymbol.owner == defn.EmptyPackageClass) Expr("")
-      else Expr(tpe.typeSymbol.owner.name)
+    def ownerNameChain(sym: Symbol): List[String] =
+      if sym.isNoSymbol then List.empty
+      else if sym == defn.EmptyPackageClass then List.empty
+      else if sym == defn.RootPackage then List.empty
+      else if sym == defn.RootClass then List.empty
+      else ownerNameChain(sym.owner) :+ normalizedName(sym)
+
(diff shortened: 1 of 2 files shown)


src/core/macro.scala
```diff
@@ -100,12 +100,13 @@ object Macro:
     import qctx.reflect._
 
     def normalizedName(s: Symbol): String = if s.flags.is(Flags.Module) then s.name.stripSuffix("$") else s.name
+    def normalizedFullName(s: Symbol): String = if s.flags.is(Flags.Module) then s.fullName.stripSuffix("$") else s.fullName
```
ghostdogpr, line 103: Since this is a full path, there might be $ in other positions than the last.

Example from caliban test suite:
expected:
`caliban.TestUtils.InvalidSchemas.Interface.ClashingObject`
actual:
`caliban.TestUtils$.InvalidSchemas$.Interface.ClashingObject`

KacperFKorban: Makes sense, see if it works for You now :slightly_smiling_face:

ghostdogpr: Yep, looks good now 👍

adamw:
Thanks! Released

PR #301, opened 2021-06-01
Fix legacy branch tests

joroKr21:
The weird project structure still makes it hard to work with though


lukaszlenart:
LGTM 👍

lukaszlenart:
Thanks a lot!

PR #302, opened 2021-06-07
Use standard project structure on legacy branch

joroKr21:


lukaszlenart:
The only downside of moving files around is that we are losing option to compare files between branches (with `master`) but that's not a big problem after all ;-)

LGTM 👍

joroKr21:
> The only downside of moving files around is that we are losing option to compare files between branches (with master) but that's not a big problem after all ;-)

I guess we could do the same on master.

PR #303, opened 2021-06-07
Added isObject to SealedTrait.Subtype

### LINKED ISSUE #164: isObject for subtypes in dispatch
Would be very useful if we could find out if a subtype is an object when using subtypes in dispatch.

sksamuel:
Fixes #164

This adds `isObject` as a parameter to `Subtype`.

One thing I was not sure on, is why the fallback implicit is required in traits used for testing (I see TypeNameInfo has it). This is only required for sealed traits which have a mixture of case classes and case objects.

### DIFF
--- a/src/core/interface.scala
+++ b/src/core/interface.scala
@@ -84,6 +84,7 @@ object SealedTrait:
                (val typeInfo: TypeInfo,
                 val annotations: IArray[Any],
                 val typeAnnotations: IArray[Any],
+                val isObject: Boolean,
                 index: Int,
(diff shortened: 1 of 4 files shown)


KacperFKorban (approved):
LGTM

adamw:
Thanks!

sksamuel:
Cool. Is there an auto snapshot ?

adamw:
Snapshot - no, but there's a release being built: https://github.com/softwaremill/magnolia/actions/runs/922133541

sksamuel:
Even better :)

PR #304, opened 2021-06-09
Semi auto derivation

mbore:


src/core/macro.scala
```diff
@@ -12,6 +12,12 @@ object Macro:
   inline def paramTypeAnns[T]: List[(String, List[Any])] = ${paramTypeAnns[T]}
   inline def repeated[T]: List[(String, Boolean)] = ${repeated[T]}
   inline def typeInfo[T]: TypeInfo = ${typeInfo[T]}
+  inline def summonOption[T]: Option[T] = ${summonOptionImpl[T]}
+
+  def summonOptionImpl[T: Type](using Quotes): Expr[Option[T]] =
+    Expr.summon[T] match
```
adamw, line 18: can't this simply be `Expr.summon[T]` without the match? :)

adamw:
Maybe we could add the recursive tests here as well?

+ some info in the readme about semi/auto derivation would be nice as well

src/core/macro.scala
KacperFKorban: The generics are switched. `Expr[Option[T]]` vs `Option[Expr[T]]`

adamw: ah 🤦

adamw:
Thanks! :)

PR #317, opened 2021-06-12
Optimization - skip error messages at depth > 1

joroKr21:
Those error messages are never shown to the user.
Also move directly reentrant error before allocating constants.


lukaszlenart (approved):
LGTM 👍

Issue #310, opened 2021-06-10
Consider not using Mirrors for derivation, so that multi-level hierarchies can be supportd


adamw:
Mirrors give us a lot "for free", but I think we could do just as well without them (that is, give the same user experience & performance), while additionally supporting multi-level hierarchies

KacperFKorban:
I agree. I've been thinking about rewriting most of the derivation logic to macros for a while.

But when it comes to ditching Mirrors altogether, there is IMHO a problem with accessing subtypes of a trait. (And that is mostly why I haven't mentioned it before) But there is also a chance that I missed something.

There is also this [docs page](https://dotty.epfl.ch/docs/reference/contextual/derivation-macro.html). It still uses Mirrors, but on a bit lower level. Maybe it would work better?

adamw:
> But when it comes to ditching Mirrors altogether, there is IMHO a problem with accessing subtypes of a trait. (And that is mostly why I haven't mentioned it before) But there is also a chance that I missed something.

Of a non-sealed trait? I suppose for sealed traits / enums we can use `Symbol.children`?

If not, maybe indeed summoning mirrors and using this information level by level would work.

KacperFKorban:
Ok, so that is the case. I just missed it... repeatedly... :facepalm: 

So I think that it should be more or less straightforward then.

ghostdogpr:
Btw hopefully Mirrors should support it soon: https://github.com/lampepfl/dotty/pull/11686

PR #321, opened 2021-06-22
Mirror-less Scala 3

KacperFKorban:
Draft of mirror-less implementation for Scala 3. Probably most of the code can be simplified a lot. But should be a good start for testing if recursive types can be handled properly this way.
Related to #310


adamw:
Thanks!

I think it might turn out recursive types work as-is, but maybe we could support value classes and non-trivial sealed trait hierarchies, for which mirrors are currently not derived.

src/core/macro.scala
```diff
@@ -13,6 +13,18 @@ object Macro:
...
+
+  def isProductImpl[T: Type](using Quotes): Expr[Boolean] =
+    import quotes.reflect.*
+    val typeSymbol = TypeRepr.of[T].typeSymbol
+    Expr(typeSymbol.isClassDef && typeSymbol.flags.is(Flags.Case))
+
+  def isSumImpl[T: Type](using Quotes): Expr[Boolean] =
+    import quotes.reflect.*
+    val typeSymbol = TypeRepr.of[T].typeSymbol
+    Expr(typeSymbol.flags.is(Flags.Sealed))
```
adamw, line 27: that's probably not enough - shouldn't we check if this is a trait as well? You could write `sealed case class` I suppose, even if it doesn't make sense

src/core/magnolia.scala
```diff
@@ -91,4 +92,138 @@ trait Derivation[TypeClass[_]] extends CommonDerivation[TypeClass]:
...
-  inline given autoDerived[A](using Mirror.Of[A]): TypeClass[A] = derived
\ No newline at end of file
+  inline given autoDerived[A](using Mirror.Of[A]): TypeClass[A] = derived
+
+trait MacroDerivation[TypeClass[_]]:
+  type Typeclass[T] = TypeClass[T]
+  def split[T](ctx: SealedTrait[Typeclass, T]): Typeclass[T]
+  def join[T](ctx: CaseClass[Typeclass, T]): Typeclass[T]
+
+  transparent inline def derived[T]: Typeclass[T] =
```
adamw, line 102: does this need to be `transparent`?

src/core/magnolia.scala
```diff
@@ -91,4 +92,138 @@ trait Derivation[TypeClass[_]] extends CommonDerivation[TypeClass]:
...
+    else if isSum[T] then
+      val sealedTrait = SealedTrait[Typeclass, T](
+        typeInfo[T],
+        MacroDerivation.getSubtypes[Typeclass, T],
+        List[Any](),
+        paramTypeAnns[T]
+      )
+      split(sealedTrait)
+    else
+      // summonInline[Typeclass[T]] // <-- this is tried to be summoned even if not needed
```
adamw, line 127: probably `Expr.summon` wouldn't have this problem, as it's not inlined in the macro itself?

adamw:
FYI the base branch is now `scala3`

KacperFKorban:
Ok, so this branch should now pass at least as many tests as the [`scala3` ](https://github.com/softwaremill/magnolia/tree/scala3) branch does.
So apart from the compatibility issues and overall being rough around the edges, it works.

@adamw will you be able to find some time to take a look at the changes and give some feedback? It would be great to test it on some real-world libraries that use `magnolia` - `tapir`?

adamw:
So one thing that breaks is calling magnolia from another macro: `Expr.summon[Schema[T]]` returns a `None`.

Even in a simple case:

```scala
    case class Test1(f1: Int)
    implicitly[Schema[Test1]]
    val codec = implicitly[MultipartCodec[Test1]].codec
```

the `implicitly[Schema[Test1]]` works, while the same lookup done as part of the multipart codec derivation macro fails

KacperFKorban:
But is it something that we actually care about? if `Expr.summon[Schema[T]]` returns `None` then we just use the same implicit method (`Typeclass.derived`) but by hand this time.
The only thing I can think of that may be affected here is the inline depth. Am I missing something?

adamw:
Not sure I understand - `Expr.summon` used to work in the old implementation, returning a `Some`. It stopped with the new one (no idea why). Where would I use `Typeclass.derived`?

adamw:
This is how to reproduce, inside `magnolia/test`:

```scala
// test1.scala
package magnolia1.tests

import magnolia1.examples._

case class Data(value: String)

object Test1 {
  summon[Show[String, Data]]
  Test2.summonShow[Data]
}

// test2.scala
package magnolia1.tests

import magnolia1.examples._
import scala.quoted.*

object Test2 {
  inline def summonShow[T]: Show[String, T] = ${ summonShowImpl[T] }

  def summonShowImpl[T: Type](using Quotes): Expr[Show[String, T]] = {
    import quotes.reflect.*

    Expr.summon[Show[String, T]].getOrElse(report.throwError(s"Cannot find a Show for: ${Type.show[T]}."))
  }
}
```

The direct `summon` works, the indirect one - fails.

OlegYch:
there is a problem with current impl which i hope can be addressed as well with the rewrite
here is a showcase https://github.com/softwaremill/magnolia/pull/421
basically, since `subtypes` contains only types directly inheriting root type, all of the logic relying on `subtypes` and `choose` breaks when you introduce non-leaf subtypes
as far as i can see the rewrite also suffers from this right now
can something be done about it?

OlegYch:
actually the problem seems to be easy to fix with old implementation, i submitted a pr

KacperFKorban:
@OlegYch You obviously can flatten the hierarchy in this rewrite too, but I'm not sure if that's what we want. Don't we lose some information, when we flatten the type hierarchy like that?

I'm pretty sure that there's a way to implement `DecoderSafe` derivation in a way that will work for those hierarchies. (choosing the instance that returns `Right`)

OlegYch:
@KacperFKorban i thought so too, but it appears to be no way to make `choose` work, or enumerate all subtypes 
trying out all instances will not work since multiple instances may succeed

PR #323, opened 2021-07-01
scala.deprecated annotation is not available in field annotations

adamw:


KacperFKorban:
@adamw Turns out that `@deprecated` annotation is on the case field and not on the constructor argument (like every other field annotation). I didn't remove the `println` yet, since the annotation has a weird `@random_uuid` suffix. I wasn't sure whether to leave it or to try adding a hack to strip it. WDYT?

adamw:
I think the suffix comes from the fact that `@deprecated` is a class, not a case class, so there's no `toString` - I don't think we can do anything about that.

I wonder though, if it's possible for the same annotation to be both on the field, and on the constructor argument? Maybe we should perform some deduplication? But then, there's no test which would reproduce such scenario, so probably not, merging :)

PR #332, opened 2021-08-07
Added isEnum field to SealedTrait

### LINKED ISSUE #319: Scala 3 enums
Scala 3 enum types seem to be coming through as a SealedTrait.
Should there be some `isEnum` flag added to detect this?

Pask423:
closes #319

### DIFF
--- a/src/core/interface.scala
+++ b/src/core/interface.scala
@@ -61,7 +61,8 @@ case class SealedTrait[Typeclass[_], Type]
                       (typeInfo: TypeInfo,
                        subtypes: IArray[SealedTrait.Subtype[Typeclass, Type, _]],
                        annotations: IArray[Any],
-                       typeAnnotations: IArray[Any]) extends Serializable:
+                       typeAnnotations: IArray[Any],
+                       isEnum: Boolean = false) extends Serializable:
 
   type Subtype[S] = SealedTrait.SubtypeValue[Typeclass, Type, S]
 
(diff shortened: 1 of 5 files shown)


src/core/interface.scala
```diff
@@ -61,7 +61,8 @@ case class SealedTrait[Typeclass[_], Type]
                       (typeInfo: TypeInfo,
                        subtypes: IArray[SealedTrait.Subtype[Typeclass, Type, _]],
                        annotations: IArray[Any],
-                       typeAnnotations: IArray[Any]) extends Serializable:
+                       typeAnnotations: IArray[Any],
+                       isEnum: Boolean = false) extends Serializable:
```
KacperFKorban: The `SealedTrait` name can be misleading, it can represent both the sealed hierarchies of `sealed trait`s as well as `enum`s. To implement this issue, a small macro would be needed similar to `isObject` here: https://github.com/softwaremill/magnolia/blob/d8be357042cbf49277fbbb0dfaa4792bbef1f348/src/core/macro.scala#L22
All flags in Quotes API can be found [here](https://github.com/lampepfl/dotty/blob/6655587a00ae72c300ea1d4a1e6966a8aff80ae7/compiler/src/scala/quoted/runtime/impl/QuotesImpl.scala#L2656).
This macro should be used in `magnolia.scala` when creating `SealedTrait` instances.

KacperFKorban (changes requested):
Hi, thanks for contributing :smiley:  

I left a comment with some requested changes.

adamw:
Thanks!

src/core/interface.scala
Pask423: Thanks for help

Issue #296, opened 2021-05-19
Missing features for Scala 3


mbore:
- [ ] support for values classes 
- [ ] support for opaque types
- [ ] support for recursive types
- [ ] accessing annotations
- [ ] a simple way to pass configuration to `join` / `split` methods (e.g. from Scala 2 -  https://github.com/softwaremill/tapir/blob/master/core/src/main/scala/sttp/tapir/generic/internal/SchemaMagnoliaDerivation.scala)

KacperFKorban:
Hi, could You elaborate on `support for recursive types` and `accessing annotations`?

mbore:
Hi, sure:
* `support for recursive types`
I should actually call it recursive class. I see that, there are already defined tests for them (e.g. `GPerson`/`RPerson` and `Recursive`) but it seems that, for the following test case, the derivation doesn't work.
```Scala
    sealed trait Node
    case class Edge(id: Long, source: Node) extends Node
    case class SimpleNode(id: Long) extends Node

    test("rerucsive class") {
      val res = summon[Show[String, Node]].show(Edge(1, SimpleNode(2)))
      println(res)
    }
```
* `accessing annotations`
I see, annotations are already supported, but the following test case from `legacy` branch doesn't work
```Scala
    case class MyAnnotation(order: Int) extends StaticAnnotation
    case class MyTypeAnnotation(order: Int) extends StaticAnnotation
    
    sealed trait AttributeParent
    @MyAnnotation(0) case class Attributed(
      @MyAnnotation(1) p1: String @MyTypeAnnotation(0),
      @MyAnnotation(2) p2: Int @MyTypeAnnotation(1)
    ) extends AttributeParent @MyTypeAnnotation(2)

    test("capture attributes against params") {
      val res = summon[Show[String, Attributed]].show(Attributed("xyz", 100))
      assertEquals(res, "Attributed{MyAnnotation(0)}{MyTypeAnnotation(2)}(p1{MyAnnotation(1)}{MyTypeAnnotation(0)}=xyz,p2{MyAnnotation(2)}{MyTypeAnnotation(1)}=100)")
    }
```
diff:
```
=> Diff (- obtained, + expected)
-Attributed{MyAnnotation(0)}(p1{MyAnnotation(1)}{MyTypeAnnotation(0)}=xyz,p2{MyAnnotation(2)}{MyTypeAnnotation(1)}=100)
[...]

PR #333, opened 2021-08-11
Implemented is OpaqueType method

Pask423:
Solves opaque types task in #296

### DIFF
--- a/src/core/interface.scala
+++ b/src/core/interface.scala
@@ -35,6 +35,7 @@ abstract class CaseClass[Typeclass[_], Type]
                         (val typeInfo: TypeInfo,
                          val isObject: Boolean,
                          val isValueClass: Boolean,
+                         val isOpaqueType: Boolean,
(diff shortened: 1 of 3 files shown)


adamw:
We definitely need a test :)

KacperFKorban:
The problem with opaque types is actually a bit more complex. Magnolia for Scala 3 is pretty much just a wrapper for Mirrors (at least for now) and Mirrors aren't generated for opaque types. And I'm pretty sure that it was a design decision. This means that magnolia cannot derive instances for opaque types.
These changes seem to work as intended, but at the current state they will always result in `isOpaque = false`.
They will be very useful once #321 works, but unfortunately, I haven't had time to work on it lately.
Not sure if we should hold this PR or merge it anyway with the intent for it to work in the future. @adamw?

adamw:
I think it's better to hold off with this PR, as it would give the false impression that this feature works, while it doesn't

We should probably concentrate on macrolia then :)

Issue #326, opened 2021-07-09
Remove scala-reflect from compile-time dependencies


adamw:
It is pulled in transitively via mercator. Instead, it should be marked as provided.

Maybe it would be possible to remove the mercator dependency alltogether?

joroKr21:
I see a couple of options for removing `mercator`:
 * Roll our own `Monad` or better yet `Applicative` type class
 * Have `constructMonadic` take polymorphic `pure` and `flatMap` functions as arguments
 * Remove `constructMonadic` altogether (does it exist for Scala 3?)

KacperFKorban:
In Scala 3 we have our own `Monad`. It was supposed to be a temporary solution.

adamw:
I think having a self-contained project w/ a `Monad`/`Monadic` construct would be the most user-friendly version. There is `constructMonadic` in the scala3 version, but without a dependency on mercator (or any other dependency)

joroKr21:
That will break binary compatibility again 😞

joroKr21:
I'm not sure we can even update Mercator without breaking binary compatibility

adamw:
@joroKr21 We're changing the group id, the package, so until we release 1.0 I don't think binary compatibility is a problem? Quite the contrary, now is the time to break things, no? :)

joroKr21:
> Quite the contrary, now is the time to break things, no? :)

On the one hand yes, but on the other hand users still complain when it's broken: #266
The problem is that Magnolia is very close to the dependency root.

adamw:
@joroKr21 but that's 0.17, 1.0 will never be a drop-in replacement for 0.17, 0.16 or any other version, even if it's only because of the new package name :)

PR #334, opened 2021-08-26
Update dependencies and setup MiMa

joroKr21:
Addresses #326 by updating Mercator. We can't remove it without breaking binary compatibility. Unfortunately we might have to do that after all but people will complain for sure.


adamw:
Thank you!

Issue #297, opened 2021-05-20
Exclude params and cons from the derivation based on annotations


vigoo:
I should have opened an issue about this originally instead of discussing it immediately in a proposed implementation (https://github.com/softwaremill/magnolia/pull/270) so I wanted to move it here now to restart discussion.

## The problem

The real-world example for the requested feature came up while trying to migrate the semi-auto derived binary codecs in https://github.com/vigoo/desert/ from shapeless to magnolia. It supports marking case class fields and sealed trait constructors as transient. These transient subtrees of a data type may have types for which there is no implicit binary codec available and it cannot be automatically derived either. Even though the transient annotations can be observed from magnolia's combine/dispatch methods, as the derivation engine does not know about these application-specific exclusion rules it tries to gather typeclass instances for these members too, and fails.

An example for using these transient annotations is when serializing Akka-typed messages for remoting but using a subset of the message type for local-only communication, transferring non-serializable references (for example an akka stream or some local resources). 

In the original PR I proposed a solution and then we discussed some alternatives. I did not check yet how the feature could be implemented in the new Scala 3 version.

It would be nice to support directly this with annotations because that way `desert` could be moved to Magnolia without changing its users code but maybe some breaking changes there are still better than maintaining a hand-written shapeless and a scala3 macro version in parallel ;) 

What do you think?

deusaquilus:
Hi Magnolia folks. Have there been any thoughts on this? I'm working for a company that would like to adopt Desert to do long-term serialization (hence the need for schema migration) but the presence of Shapeless is a non-starter because of compile-time issues. Could this issue be solved in magnolia on a lower level?

adamw:
I don't think there has been any developments around this, but we're open to discussion and contributions :) I guess the main question is what would happen when calling `caseClass.param.typeclass` for a transient field? An exception would be thrown?

deusaquilus:
I think that's reasonable. Just like calling .get on a `None` or a `Failure`.

adamw:
Ok, well I don't see a reason not to support such transient fields. We just have to come up with a good name (`@deriveTransient`?) and implement :)

vigoo:
@adamw so you suggest to have a built-in `@deriveTransient` (or whatever) annotation for this. I think that would work and probably the simplest approach.

Two things come to my mind:
- It should work both on fields and constructors 
[...]

PR #337, opened 2021-09-14
Methods rename

### LINKED ISSUE #247: rename combine/dispatch to join/split
These methods are quite fundamental, so this will be a trivial, but breaking change. I would like to make this change before v1.0, though.

There is a possibility of having a transition period where we allow the methods to be called either `join` or `combine` and `split` or `dispatch`.

Pask423:
closes #247

### DIFF
--- a/core/src/main/scala/magnolia1/interface.scala
+++ b/core/src/main/scala/magnolia1/interface.scala
@@ -424,7 +424,7 @@ final class SealedTrait[Typeclass[_], Type](
     *  @return  the result of applying the `handle` lambda to subtype of the sealed trait which
     *           matches the parameter `value`
     */
-  def dispatch[Return](value: Type)(handle: Subtype[Typeclass, Type] => Return): Return = {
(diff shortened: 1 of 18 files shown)


KacperFKorban:
Thanks for the contribution!

When it comes to changing the names, I think that we should deprecate the old names first.
But I don't know much about scala2 macros, so I'm not sure how hard it would be. @adamw

adamw:
@KacperFKorban ideally - yes - we would want to support both names with a deprecation notice. However, given that our resources are limited, and I'd like to get magnolia 1.0 (for both Scalas) in finite time, I think it would be better to do the renaming so that both versions use the same vocabulary, rather than keeping the status quo. There's still two major changes that should come in for the Scala 2 version (dropping mercator & possibly https://github.com/softwaremill/magnolia/issues/297 - cc @joroKr21 @deusaquilus) before releasing 1.0 anyway.

But of course that's only my opinion, if others would have a different one, that's fine :)

joroKr21:
Yeah I think a deprecation cycle might not be feasible, given that it would be just another milestone.

PR #338, opened 2021-09-19
Mercator removal

### LINKED ISSUE #326: Remove scala-reflect from compile-time dependencies

Pask423:
Should close #326


lukaszlenart:
The build fails on _binary incompatibility_ check but I think that's ok.

core/src/main/scala/magnolia1/monadic.scala
```diff
@@ -0,0 +1,56 @@
...
+  type Apply[X] = F[X]
+  def point[A](value: A): F[A]
+  def map[A, B](from: F[A])(fn: A => B): F[B]
+}
+
+trait Monadic[F[_]] extends Functor[F] {
+  def flatMap[A, B](from: F[A])(fn: A => F[B]): F[B]
+}
+
+object Monadic {
```
adamw, line 13: I assume this is copied from mercator, we should add a comment pointing back to the original source

core/src/main/scala/magnolia1/monadic.scala
```diff
@@ -0,0 +1,56 @@
+package magnolia1
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Success, Try}
+
+trait Functor[F[_]] {
```
adamw: do we need the `Functor` as a stand-alone abstraction anywhere? My worry would be that it's such a common name, that we would add to the import-pollution when magnolia is on the classpath. Maybe we could simply have only `Monadic` with all the methods?

core/src/main/scala/magnolia1/monadic.scala
Pask423: As I wrote in commit msg it was copied from scala3 branch so it depends on from where it was copied there

adamw: Ah :D I didn't read the commit msg, that's fine then

joroKr21:
I would advise to not expose the instances as objects / classes in the interest of binary compatibility for the future.

adamw:
@joroKr21 you mean expose them as `def`s instead? I didn't know that this has implications for bin compat, but I guess it makes sense, as a `def` gives you much more flexiblity

joroKr21:
> @joroKr21 you mean expose them as defs instead? I didn't know that this has implications for bin compat, but I guess it makes sense, as a def gives you much more flexiblity

Yes, exactly - well if the object/class is public you can't remove it in a binary compatible way. And the would also be reflected in the types of the instances so you wouldn't be able to change that either.

adamw:
@joroKr21 Thanks, makes sense :) We have to do the same change on the scala3 branch as well then

Pask423:
Ok I will do it for scala2 but I think we need separate issue for scala3

adamw:
@Pask423 yeah a separate PR for sure :)

PR #347, opened 2021-10-05
Wrong typeInfo for enum case values

### LINKED ISSUE #346: Wrong typeInfo for enum case values
Yeah that's a problem with magnolia. Given:

```scala
enum Size:
  case S, M, L

  test("construct a Show instance for an enum") {
    val res = Show.derived[Size].show(Size.S)
    assertEquals(res, "S")
  }
```

the result is `Size`, not `S`. The child typeclasses are incorrect.

_Originally posted by @adamw in https://github.com/softwaremill/magnolia/issues/345#issuecomment-934335024_

KacperFKorban:
fixes #346

### DIFF
--- a/src/core/macro.scala
+++ b/src/core/macro.scala
@@ -167,9 +167,11 @@ object Macro:
   def typeInfo[T: Type](using Quotes): Expr[TypeInfo] =
     import quotes.reflect._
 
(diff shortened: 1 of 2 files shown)


adamw:
Thanks!

PR #351, opened 2021-10-08
Replace summonOption with summonFrom

joroKr21:
`summonOption` generates both branches and resolves at runtime.
`summonFrom` resolves at compile time and generates only one branch.


KacperFKorban (approved):

adamw:
Thanks! :)

PR #376, opened 2022-01-27
Inherit annotations

### LINKED ISSUE #375: allow inheriting annotations from base trait/classes
* add `@inherit` annotation working on def's
* extend collecting annotations in magnolia based on that -> `annotationsOf(Symbol)`

kubinio123:


core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -181,8 +181,24 @@ object Magnolia {
...
-    def annotationsOf(symbol: Symbol): List[Tree] =
-      annotationTrees(symbol.annotations)
+    def annotationsOf(symbol: Symbol): List[Tree] = {
+      @tailrec
+      def fromBaseClasses(owner: Symbol): List[Annotation] =
+        if (owner.isClass) {
+          owner.asClass.baseClasses
+            .flatMap(_.asType.toType.members)
+            .collect {
+              case t: TermSymbol if t.name == symbol.name && t.annotations.exists(isInherit) => t.annotations.filterNot(isInherit)
```
kubinio123: I wanted to compare `Symbols` with `==` but that didn't work so I compare just by `name` but if feels risky to do just that

adamw: hm I think if we're looking only at the symbols defined in super-classes / super-traits that's fine. Although, if the symbol is a method, we might also compare the number or aguments or agument types. That will be an exact match. Maybe there's a way to obtain the method's signature and compare that?

adamw: although, signatures aren't 100% fail-safe either, as an overriding method may specialise parameters. But since we are looking at no-arg defs anyway, probably just comparing the name & arity is good enough :)

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -181,8 +181,33 @@ object Magnolia {
           annotation.tree
       }
 
-    def annotationsOf(symbol: Symbol): List[Tree] =
-      annotationTrees(symbol.annotations)
+    def annotationsOf(symbol: Symbol): List[Tree] = {
+      @tailrec
+      def fromBaseClassesMembers(owner: Symbol): List[Annotation] =
+        if (owner.isClass) {
+          owner.asClass.baseClasses
```
adamw, line 188: it might be useful to filter out classes which for sure won't have any annotations, e.g. `java.lang.Object` or any traits from `scala.*`

core/src/main/scala/magnolia1/interface.scala
```diff
@@ -476,6 +476,11 @@ final case class TypeName(owner: String, short: String, typeArguments: Seq[TypeN
   */
 final class debug(typeNamePart: String = "") extends scala.annotation.StaticAnnotation
 
+/** This annotation can be attached on the contents of base coproduct (trait/class) along with another specific annotations we want to inherit.
+  * It causes magnolia to collect annotations from base class members.
+  */
+final class inherit extends scala.annotation.StaticAnnotation
```
adamw: I'm wondering what's the best approach here.

(a) as implemented now - use `@inherit` at usage-site to specify that annotations in that particular case should be inherited
(b) use `@inherit` as a meta-annotation to specify that a given annotation (such as tapir's `@description` should always be inherited
(c) add a separate `inheritedAnnotations` method to `Param` / `CaseClass` so that the derivation method might use the inherited annotations, or might skip them

@seglo maybe you would have some opinion here as well?

One issue to consider is what to do when there are annotations both on the implementation and the definition. Probably the one on the implementation should take precedence, and it will be the first on the list, right? We should specify that in the docs so it's clear what to expect.

kubinio123: I like option b) and c) since users would not have to use `@inherit` in their code. 
Currently in case of annotation being present in definition and implementation the second will take precedence - since we first take `symbol.annotations` then add inherited and `distinct` at the end - but I will add a test for that and docs as well.

seglo: > (b) use `@inherit` as a meta-annotation to specify that a given annotation (such as tapir's `@description` should always be inherited

I'm not sure what you mean by meta-annotation, exactly, but I like the idea of not having to enable this type of inheritance at usage sites. The intuition that annotations should be inherited is pretty common and generally already happens at runtime in Java and Scala (depending on the base annotation type) so I think inheritance in this case should work the same.

I don't have a resource handy, but I recall reading that annotations are only inherited from base `class` definitions and not traits in Scala. That restriction makes the rules a little easier to define since you don't have to worry about conflicts between base annotations when your type inherits multiple traits that have the same base type and annotation. A class can only be inherited once.

> One issue to consider is what to do when there are annotations both on the implementation and the definition. Probably the one on the implementation should take precedence, and it will be the first on the list, right? We should specify that in the docs so it's clear what to expect.

Agreed.

seglo: > I don't have a resource handy, but I recall reading that annotations are only inherited from base `class` definitions and not traits in Scala. That restriction makes the rules a little easier to define since you don't have to worry about conflicts between base annotations when your type inherits multiple traits that have the same base type and annotation. A class can only be inherited once.

That said it would be nice to have a way of inheriting from traits too to make OpenAPI schema's more composable in the code :)

kubinio123: This works for `trait`'s as well :)

adamw: @seglo a meta-annotation is an annotation on another annotation, that is we would annotate `@description` with `@inherit` specifying that it is inheritable. But that said, in fact we would annotate all of tapir's annotations like that, so this brings the question, maybe this should be the default?

seglo: Got it. My comments were from a user PoV in tapir. I think there tapir's OpenAPI annotations should inherit by default, but I'll defer an opinion on what's best for Magnolia's API since I don't have much experience with it.

adamw: I think I'm leaning towards (b), though (c) has the upside that it preserves more information, and allows user-defined disambiguation in case an annotation is defined both on the definition and the implementation. Hard choice ;)

kubinio123: I will go with c) since as you said earlier we would make all annotations in tapir `@inherit`  anyway. Also with c) one can handle more cases than we had discussed like maybe someone will want to "merge" annotation from definition with implementation if it carries some value

adamw: Ok let's try :) we'll see how much impact it has and if MiMa doesn't shout at us ;)

adamw:
Great! :) Let's see if we can do something in scala3 as well

PR #378, opened 2022-01-31
Inherit annotations s3

### LINKED ISSUE #375: allow inheriting annotations from base trait/classes

kubinio123:


src/core/macro.scala
```diff
@@ -193,3 +144,99 @@ object Macro:
...
+        tpe.typeSymbol.annotations
+          .filter(filterAnnotation)
+          .map(_.asExpr.asInstanceOf[Expr[Any]])
+      }
+
+    def inheritedAnns: Expr[List[Any]] =
+      Expr.ofList {
+        tpe.baseClasses
+          .collect {
+            case s if s.name != tpe.typeSymbol.name => s.annotations
```
KacperFKorban: Comparing names is not 100% reliable, wouldn't comparing type symbols work?

src/core/macro.scala
```diff
@@ -193,3 +144,99 @@ object Macro:
...
+          } // skip self
+          .flatten
+          .filter(filterAnnotation)
+          .map(_.asExpr.asInstanceOf[Expr[Any]])
+      }
+
+    def typeAnns: Expr[List[Any]] =
```
KacperFKorban: Could you add an end marker or curly braces to this def?

src/core/macro.scala
```diff
@@ -193,3 +144,99 @@ object Macro:
...
+          (fromConstructor(tpe.typeSymbol) ++ fromDeclarations(tpe.typeSymbol))
+            .filter { case (_, anns) => anns.nonEmpty }
+        }
+      }
+
+    def inheritedParamAnns: Expr[List[(String, List[Any])]] =
+      Expr.ofList {
+        groupByParamName {
+          tpe.baseClasses
+            .filterNot(bc =>
```
KacperFKorban: Could you switch to curly braces here?

src/core/interface.scala
```diff
@@ -1,9 +1,10 @@
 package magnolia1
 
-import language.experimental.macros
 import scala.annotation.tailrec
 import scala.reflect.*
 
+import language.experimental.macros
```
KacperFKorban: Does this import do anything anymore in Scala 3?

src/core/macro.scala
kubinio123: Thanks, seems to work when I just compare `s` against `tpe.typeSymbol`

src/core/interface.scala
kubinio123: I don't think so, not sure how it end up here :D

KacperFKorban (approved):
LGTM, but I would wait for @adamw approval as well :smile:

src/examples/show.scala
```diff
@@ -66,13 +70,21 @@ trait GenericShow[Out] extends AutoDerivation[[X] =>> Show[Out, X]] {
...
         val annotationStr =
           if (anns.isEmpty) "" else anns.mkString("{", ",", "}")
 
         prefix(annotationStr, sub.typeclass.show(sub.value))
       }
+
+  def filterUnwantedAnns(anns: IArray[Any]): IArray[Any] =
+    anns
+      .filterNot(_.isInstanceOf[scala.SerialVersionUID])
+      .filterNot(_.isInstanceOf[scala.annotation.showAsInfix])
```
adamw: shouldn't we filter out these annotations from `inheritedAnns` in the first place?

src/core/magnolia.scala
```diff
@@ -69,6 +72,39 @@ trait CommonDerivation[TypeClass[_]]:
...
+        ) ::
+          getParams_[T, ltail, ptail](
+            annotations,
+            inheritedAnnotations,
+            typeAnnotations,
+            repeated,
+            idx + 1
+          )
+
+  // for backward compatibility with v1.0.0
```
adamw, line 107: can this impl call getParams_ to avoid code duplication?

src/core/interface.scala
```diff
@@ -74,6 +118,31 @@ abstract class CaseClass[Typeclass[_], Type](
...
+      typeAnnotations
+    ):
+      type PType = P
+      def default: Option[PType] = defaultVal.value
+      def typeclass = cbn.value
+      override def inheritedAnnotations = inheritedAnns
+      def deref(value: Type): P =
+        value.asInstanceOf[Product].productElement(idx).asInstanceOf[P]
+
+  // for backward compatibility with v1.0.0
```
adamw, line 145: same here, I'd just call param with the default inhertiedAnns value

src/core/macro.scala
```diff
@@ -193,3 +144,100 @@ object Macro:
...
+    def anns: Expr[List[Any]] =
+      Expr.ofList {
+        tpe.typeSymbol.annotations
+          .filter(filterAnnotation)
+          .map(_.asExpr.asInstanceOf[Expr[Any]])
+      }
+
+    def inheritedAnns: Expr[List[Any]] =
+      Expr.ofList {
+        tpe.baseClasses
```
adamw, line 162: maybe we should filter the base classes here as well?>

src/examples/show.scala
kubinio123: yep, after filtering out `Object` and `scala*` we don't need to filter these annotations

src/core/magnolia.scala
kubinio123: yes, missed that some how

adamw:
Thanks @kubinio123 @KacperFKorban  :)

Issue #36, opened 2017-11-11
Support Scala Native


propensive:
It would be nice if Magnolia worked on Scala Native. Anyone familiar with setting up Scala Native in SBT could probably tackle this.

xuwei-k:
scala-native does not support sbt 1.x
- https://github.com/scala-native/scala-native/pull/1039
- https://github.com/propensive/magnolia/blob/eb53a6df0/project/build.properties

propensive:
Aha, knowing that is going to same me some time, @xuwei-k -- thanks. I've been planning to build Scala Native programs directly from the command line (without SBT), but couldn't find enough documentation. Maybe I'll try a bit harder to find out, and I can do it this way.

mielientiev:
also scala native doesn't support scala 2.12
- https://github.com/scala-native/scala-native/issues/233

propensive:
We're still supporting 2.11.x, so at least that shouldn't be too big a problem. :)

PR #389, opened 2022-03-22
Support Scala native (Scala 2)

adampauls:
I think this is as easy as it looks? Fixes https://github.com/softwaremill/magnolia/issues/36.


adamw:
Thanks!

Issue #242, opened 2020-05-04
Spurious compilation errors


OlegYch:
With typeclass like this
```
trait MagnoliaFormatSchema[T] extends MagnoliaFormatSchema.Input[T]
object MagnoliaFormatSchema {
  trait Input[T] 
  type Typeclass[T] = MagnoliaFormatSchema[T]
  trait Derived[T] extends Typeclass[T] 

  def combine[T](ctx: CaseClass[Typeclass, T]): Derived[T] = {
    ???
  }

  def dispatch[T](ctx: SealedTrait[Typeclass, T]): Derived[T] = {
    ???
  }

  def gen[T]: Derived[T] = macro Magnolia.gen[T]
}
```
and case class like this
```
object ObjectName {
  case class Step(
      tpe: ObjectName.Type.Value,
      code: Step.Id,
      order: Int,
      action: Step.Action.Value,
      name: String,
      auto: Boolean,
      autoCheck: Boolean,
      headline: Option[String],
      rollbackCode: Option[Step.Id],
  )
}
```
i sometimes get this:
```
 type mismatch;
[error]  found   : MagnoliaFormatSchema.Typeclass[packagename.ObjectName.Step]
[error]     (which expands to)  MagnoliaFormatSchema[packagename.ObjectName.Step]
[error]  required: MagnoliaFormatSchema.Derived[packagename.ObjectName.Step]
[error]   implicit val zStepFormat                 = MagnoliaFormatSchema.gen[Step]
```
and sometimes this
```
[warn] Generated class packagename.ObjectName$$typecreator2$1 differs only in case from packagename.ObjectName$$typecreator2$1.
[warn]   Such classes will overwrite one another on case-insensitive filesystems.
```
other times it compiles just fine
using magnolia 0.16.0 with scala 2.13.2 on windows
sorry i can't minimize it properly because it's very spurious

OlegYch:
adding explicit type to `zStepFormat` seems to sidestep it

joroKr21:
Spurious issue means it's probably cause by caching.
And indeed when I think about it, caching assumes `Typeclass[T]` is the type returned from `gen`. This kind of usage breaks the assumption and perhaps causes spurious issues.

I'm not sure about the `typecreator` name clash. Sounds like something specific to `ObjectName.Type.Value`.

OlegYch:
i can see in debug output that one of the params is wrapped in 'Deferred' when it fails, like this:
```
[...]

PR #399, opened 2022-05-10
Refactor Derivation interface

OlegYch:
Move implementation out of Derivation for easier reuse in case a different interface is needed

### DIFF
--- a/src/core/impl.scala
+++ b/src/core/impl.scala
@@ -99,12 +99,14 @@ object CaseClassDerivation:
           )
 end CaseClassDerivation
 
-trait SealedTraitDerivation[TypeClass[_]]:
-  protected inline def deriveSubtype[s](m: Mirror.Of[s]): TypeClass[s]
+trait SealedTraitDerivation:
+  type Typeclass[T]
+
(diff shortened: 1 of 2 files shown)


OlegYch:
@adamw can this be released please?

adamw:
@OlegYch sorry for the delay. I'm hesitant as this would increase the surface area of binary compatibility that we have to maintain quite significantly (as it would include the inline method implementations).

OlegYch:
@adamw i agree, but these are methods already exposed in public api, just packaged in a more a reusable way

adamw:
You are right, it's already public :) I've adjusted the naming abit, let me know what you think

OlegYch:
@adamw naming changes look good, but promoting type member to type parameter makes it less flexible and wouldn't work for my specific usecase (sample here https://github.com/softwaremill/magnolia/issues/242 )

adamw:
@OlegYch now? :)

OlegYch:
@adamw LGTM

adamw:
Should be released soon :)

PR #415, opened 2022-08-08
Fallback for accessing properties on symbols of proxy types

KacperFKorban:
Looks like the crash can only happen with `case val`s, so a simple fallback will work. (Those types don't have any params either way)


adamw:
Awesome, thanks! :)

adamw:
This closes #413

PR #418, opened 2022-08-19
Readme fix for Scala2

alexandru:
The `Typeclass` type alias is confusing, as it's unused. When I read the documentation, I wondered if that type alias is needed by the macro or something 🙂

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -47,8 +47,6 @@ The derivation typeclass for a `Show` typeclass might look like this:
 import language.experimental.macros, magnolia1._
 
 object ShowDerivation {
-  type Typeclass[T] = Show[T]
-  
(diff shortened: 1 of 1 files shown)


KacperFKorban (approved):

adamw:
It is actually used and required:  https://github.com/softwaremill/magnolia/blob/scala2/core/src/main/scala/magnolia1/magnolia.scala#L69

If you try to remove the `Typeclass` type alias from the derivation it will fail (e.g. in the tests). I suspect the motivation might be supporting types with more than one type parameter, so that magnolia knows which parameter to vary. Not sure if it could be inferred.

alexandru:
Interesting. In my own code I removed the alias without consequences, not sure what happened there.

I'll double check, but will close this PR for now.

PR #419, opened 2022-08-19
Readme fix for Scala3

alexandru:
Sample is using an undefined type alias.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -43,7 +43,7 @@ trait Print[T] {
 }
 
 object Print extends AutoDerivation[Print]:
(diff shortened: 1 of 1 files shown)


KacperFKorban (approved):

adamw:
The type is actually defined here: https://github.com/softwaremill/magnolia/blob/scala3/src/core/magnolia.scala#L10

Though unlike for scala2, here it doesn't need to be user-defined and using the concrete type works just well. Which is definitely more user-friendly :)

adamw:
Thanks!

PR #421, opened 2022-08-29
Fix getting subtypes when hierarchy has intermediate types

OlegYch:


adamw:
Great!

Could you add a separate test, without reusing the existing ones? Would be good to have both cases (simple & hierarchy) covered

OlegYch:
@adamw added
one thing to note that i had to replace using `Sum.ordinal` with simple `isInstanceOf` check
not sure why it wasn't done initially like so, perhaps of scala.js compatibility? the tests pass fine though...

src/core/impl.scala
```diff
@@ -118,24 +118,32 @@ trait SealedTraitDerivation:
...
+                IArray.from(anns[s]),
+                IArray.from(inheritedAnns[s]),
+                IArray.from(paramTypeAnns[A]),
+                isObject[s],
+                idx,
+                CallByNeed(summonFrom {
+                  case tc: Typeclass[`s`] => tc
+                  case _ => deriveSubtype(summonInline[Mirror.Of[s]])
+                }),
+                x => x.isInstanceOf[s & A],
```
adamw, line 143: will this work for `enum`s, though? If they don't have parameters they are values, not types, so the typecheck won't work.

adamw: question is - what are the ordinals in such complex trees? maybe we should be checking ordinal + parent?

src/core/impl.scala
```diff
@@ -118,24 +118,32 @@ trait SealedTraitDerivation:
...
-          idx,
-          CallByNeed(summonFrom {
-            case tc: Typeclass[`s`] => tc
-            case _                  => deriveSubtype(summonInline[Mirror.Of[s]])
-          }),
-          x => m.ordinal(x) == idx,
-          _.asInstanceOf[s & A]
-        ) :: subtypesFromMirror[A, tail](m, idx + 1)
+        val sub = summonFrom {
+          case mm: Mirror.SumOf[`s`] =>
```
adamw, line 128: I don't quite understand what didn't work before? I guess if the subtype is itself a sum, it would get derived by the recursive call to `deriveSubtype`?

src/core/impl.scala
OlegYch: apparently it works for enums as well, added a test for that

src/core/impl.scala
OlegYch: yes, it would get derived, but the subtypes will not be in `subtypes` and there is no way to recover them

src/core/impl.scala
adamw: thanks! Indeed, typechecks using `.isInstanceOf[Size.S.type]` work - not sure how that works behind the scenes, @KacperFKorban do you know maybe? (just out of curiosity)

src/core/impl.scala
adamw: hm, anyway for complex hierarchies which don't have to form a nice tree that's probably the best we can do anyway; thanks :)

adamw:
Thanks! :)

src/core/impl.scala
KacperFKorban: I think that it is just compiled to reference equality. (`if_acmpeq` or `if_acmpne`)

adamw: ah makes sense, thanks :)

PR #431, opened 2022-11-02
Tests refactor

micsza:


src/test/SumsTests.scala
```diff
@@ -0,0 +1,229 @@
...
+  case object Pink extends Color
+
+  sealed trait Y
+  case object A extends Y
+  case class B(s: String) extends Y
+
+  enum Size:
+    case S, M, L
+
+  @MyAnnotation(0)
```
adamw: do we need the annotations in theses tests?

adamw:
Great to have these categorized, thanks! :)

PR #435, opened 2022-11-22
Support for value classes

micsza:
Implements support for mirrorless value classes derivation for non-generic products with annotations and default values.

TODOs:
1. support for non-product value classes
2. support for generic value classes
3. support for access modifiers

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -74,6 +74,7 @@ lazy val test = (projectMatrix in file(".test"))
   .settings(
     name := "magnolia-test",
     scalacOptions += "-Yretain-trees",
+    scalacOptions += "-Xcheck-macros", // TODO remove
     projectDependencies ++= Seq(
       "org.scalameta" %%% "munit" % "1.0.0-M6"
     ),
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import com.softwaremill.UpdateVersionInDocs
 import com.softwaremill.SbtSoftwareMillCommon.commonSmlBuildSettings
 import com.softwaremill.Publish.{updateDocs, ossPublishSettings}
 
-val scala3 = "3.2.0"
+val scala3 = "3.2.1"
 
 ThisBuild / dynverTagPrefix := "scala3-v" // a custom prefix is needed to differentiate tags between scala2 & scala3 versions
 
--- a/build.sbt
+++ b/build.sbt
@@ -74,7 +74,6 @@ lazy val test = (projectMatrix in file(".test"))
   .settings(
     name := "magnolia-test",
     scalacOptions += "-Yretain-trees",
-    scalacOptions += "-Xcheck-macros", // TODO remove
     projectDependencies ++= Seq(
       "org.scalameta" %%% "munit" % "1.0.0-M6"
     ),
--- a/src/core/impl.scala
+++ b/src/core/impl.scala
@@ -7,6 +7,31 @@ import scala.reflect.*
 import Macro.*
 
 object CaseClassDerivation:
+
+
+  inline def fromMacro[TC[_], A]: CaseClass[TC, A] = valueClassDerivation[TC, A]
+
+  inline def valueClassDerivation[TC[_], A]: CaseClass[TC, A] =
+
+    val params: IArray[CaseClass.Param[TC, A]] = ???
+
+    new CaseClass[TC, A](
+      typeInfo = typeInfo[A],
+      isObject = isObject[A],
+      isValueClass = isValueClass[A],
+      params = params,
+      annotations = IArray.from(anns[A]),
+      inheritedAnnotations = IArray.from(inheritedAnns[A]),
+      typeAnnotations = IArray.from(typeAnns[A])
+    ):
+      def construct[PType: ClassTag](makeParam: Param => PType): A = ???
+
+      def rawConstruct(fieldValues: Seq[Any]): A = ???
+
+      def constructEither[Err, PType: ClassTag](makeParam: Param => Either[Err, PType]): Either[List[Err], A] = ???
+
+      def constructMonadic[M[_]: Monadic, PType: ClassTag](makeParam: Param => M[PType]): M[A] = ???
+
   inline def fromMirror[Typeclass[_], A](
       product: Mirror.ProductOf[A]
   ): CaseClass[Typeclass, A] =
--- a/src/core/impl.scala
+++ b/src/core/impl.scala
@@ -5,32 +5,51 @@ import scala.deriving.Mirror
 import scala.reflect.*
 
 import Macro.*
+import MirrorlessMacro.*
 
 object CaseClassDerivation:
 
+  inline def valueClassDerivation[TC[_], V]: CaseClass[TC, V] =
 
-  inline def fromMacro[TC[_], A]: CaseClass[TC, A] = valueClassDerivation[TC, A]
+    val theParam: CaseClass.Param[TC, V] = ValueClassDerivation.getValueClassParam[TC, V]
 
-  inline def valueClassDerivation[TC[_], A]: CaseClass[TC, A] =
-
-    val param: CaseClass.Param[TC, A] = getValueClassParam[TC, A]
-
-    new CaseClass[TC, A](
-      typeInfo = typeInfo[A],
-      isObject = isObject[A],
-      isValueClass = isValueClass[A], // just true?
-      params = IArray.from(param :: Nil),
-      annotations = IArray.from(anns[A]),
-      inheritedAnnotations = IArray.from(inheritedAnns[A]),
-      typeAnnotations = IArray.from(typeAnns[A])
+    new CaseClass[TC, V](
+      typeInfo = typeInfo[V],
+      isObject = isObject[V],
+      isValueClass = isValueClass[V], // just true?
+      params = IArray.from(theParam :: Nil),
+      annotations = IArray.from(anns[V]),
+      inheritedAnnotations = IArray.from(inheritedAnns[V]),
+      typeAnnotations = IArray.from(typeAnns[V])
     ):
-      def construct[PType: ClassTag](makeParam: Param => PType): A = ???
 
-      def rawConstruct(fieldValues: Seq[Any]): A = ???
+      def rawConstruct(fieldValues: Seq[Any]): V = 
+        ValueClassDerivation.rawContructByMacro[V](fieldValues.toList)
+    
+      def construct[PType: ClassTag](makeParam: Param => PType): V = 
+        ValueClassDerivation.rawContructByMacro[V](params.map(makeParam).toList)
+    
+      def constructEither[Err, PType: ClassTag](makeParam: Param => Either[Err, PType]): Either[List[Err], V] = 
+        params
+        .map(makeParam)
+        .foldLeft[Either[List[Err], Array[PType]]](Right(Array())) { 
+          case (Left(errs), Left(err)) => Left(errs ++ List(err))
+          case (Right(acc), Right(param)) => Right(acc ++ Array(param))
+          case (errs @ Left(_), _)        => errs
+          case (_, Left(err))             => Left(List(err))
+        }
+        .map { params => ValueClassDerivation.rawContructByMacro[V](params.toList) }
 
-      def constructEither[Err, PType: ClassTag](makeParam: Param => Either[Err, PType]): Either[List[Err], A] = ???
+      def constructMonadic[M[_]: Monadic, PType: ClassTag](makeParam: Param => M[PType]): M[V] = 
+        val m = summon[Monadic[M]]
+        m.map {
+          params.map(makeParam).foldLeft(m.point(Array())) { (accM, paramM) =>
+            m.flatMap(accM) { acc =>
+              m.map(paramM)(acc ++ List(_))
+            }
+          }
+        } { params => ValueClassDerivation.rawContructByMacro[V](params.toList) }
 
-      def constructMonadic[M[_]: Monadic, PType: ClassTag](makeParam: Param => M[PType]): M[A] = ???
 
   inline def fromMirror[Typeclass[_], A](
       product: Mirror.ProductOf[A]
@@ -110,9 +129,9 @@ object CaseClassDerivation:
           repeated.getOrElse(label, false),
           CallByNeed(summonInline[Typeclass[p]]),
           CallByNeed(defaults.get(label).flatten.map(_.apply.asInstanceOf[p])),
-          IArray.from(annotations.getOrElse(label, List())),
-          IArray.from(inheritedAnnotations.getOrElse(label, List())),
-          IArray.from(typeAnnotations.getOrElse(label, List()))
+          annotations.getOrElse(label, List()),
+          inheritedAnnotations.getOrElse(label, List()),
+          typeAnnotations.getOrElse(label, List())
         ) ::
           paramsFromMaps[Typeclass, A, ltail, ptail](
             annotations,
@@ -122,6 +141,7 @@ object CaseClassDerivation:
             defaults,
             idx + 1
           )
+
 end CaseClassDerivation
 
 trait SealedTraitDerivation:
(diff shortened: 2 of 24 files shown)


src/core/interface.scala
```diff
@@ -33,6 +34,28 @@ object CaseClass:
...
 
   object Param:
+
+    def apply[F[_], T, P](
+        name: String,
+        idx: Int,
+        repeated: Boolean,
+        cbn: CallByNeed[F[P]],
+        defaultVal: CallByNeed[Option[P]],
+        annotations: List[Any],
```
adamw: does this only differ in using `List` instead of `IArray`? If so, can we keep a single variant? that way the public api (the one for which we need to keep compatibility) will be smaller

src/core/magnolia.scala
```diff
@@ -77,12 +79,27 @@ trait Derivation[TypeClass[_]]
       case sum: Mirror.SumOf[A]         => derivedMirrorSum[A](sum)
       case product: Mirror.ProductOf[A] => derivedMirrorProduct[A](product)
 
-  inline def derived[A](using Mirror.Of[A]): Typeclass[A] = derivedMirror[A]
+  inline def derivedValueClass[A]: Typeclass[A] = join(CaseClassDerivation.valueClassDerivation)
```
adamw: can these be private?

src/core/magnolia.scala
```diff
@@ -77,12 +79,27 @@ trait Derivation[TypeClass[_]]
...
+
+  inline def derivedMirrorless[A]: Typeclass[A] = 
+    inline if Macro.isValueClass[A] then 
+      derivedValueClass[A]
+    else error("Mirrrorful derivation not posiible, mirrorless derivation either not possible or not implemented yet.")
```
adamw: typos :) plus I would make this a bit more user-friendly, sth like `Deriving the typeclass based on mirrors or directly is not possible for $A. Please refer to the documentation or report a feature request` (it's important to include the type name as this can be a nested invocation)

src/core/impl.scala
```diff
@@ -5,8 +5,52 @@ import scala.deriving.Mirror
...
 object CaseClassDerivation:
+
+  inline def valueClassDerivation[TC[_], V]: CaseClass[TC, V] =
+
+    val theParam: CaseClass.Param[TC, V] = ValueClassDerivation.getValueClassParam[TC, V]
+
+    new CaseClass[TC, V](
+      typeInfo = typeInfo[V],
+      isObject = isObject[V],
+      isValueClass = isValueClass[V], // just true?
```
adamw: I guess so, just true ;)

build.sbt
```diff
@@ -74,6 +74,7 @@ lazy val test = (projectMatrix in file(".test"))
   .settings(
     name := "magnolia-test",
     scalacOptions += "-Yretain-trees",
+    scalacOptions += "-Xcheck-macros", // TODO remove
```
adamw: todo? :)

src/test/ValueClassesTests.scala
```diff
@@ -2,13 +2,63 @@ package magnolia1.tests
...
+    val res = Print.derived[WithDefault].print(WithDefault())
+    assertEquals(res, "WithDefault(123)")
+  }
+
+  test("Semi default") {
+    val res = SemiDefault.derived[WithDefault].default
+    assertEquals(res, WithDefault(123))
+  }
+
   // test("serialize a value class") {
```
adamw, line 102: what are we missing to enable these tests?

adamw: ah ok I see above :)

adamw: maybe add a comment fo reach test why it fails?

src/test/SumsTests.scala
```diff
@@ -96,25 +96,30 @@ class SumsTests extends munit.FunSuite:
...
+    //  mirrorful approach failed with [...]
+    //  mirrorless approach failed with [...]
+    //
+    // assert(
+    //   error contains "No given instance of type deriving.Mirror.Of[magnolia1.tests.SumsTests.Parent] was found for parameter x$1 of method derived in trait Derivation."
+    // )
+    // assert(
+    //   error contains "trait Parent is not a generic sum because its child trait BadChild is not a generic product because it is not a case class"
+    // )
+    assert(error contains "Mirrrorful derivation not posiible, mirrorless derivation either not possible or not implemented yet.")
```
adamw: at the very least we need to retain the type for which things fail

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
 
     private def isObjectOrScala(bc: Symbol) =
       bc.name.contains("java.lang.Object") || bc.fullName.startsWith("scala.")
 
     private def filterAnnotation(a: Term): Boolean =
       a.tpe.typeSymbol.maybeOwner.isNoSymbol ||
         a.tpe.typeSymbol.owner.fullName != "scala.annotation.internal"
-  }
+
+object MirrorlessMacro:
```
adamw: aren't the other macros mirrorless as well? maybe let's just group by functionality, `ValueClassDerivation` seems good :)

plus, can we make it private? I know the others are not, they should probably be, but let's not increase the API surface area if we don't have to

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+      ${ rawContructByMacroImpl[V]('as) } 
+
+    def getValueClassParamImpl[TC[_]: Type, A: Type](using Quotes): Expr[CaseClass.Param[TC, A]] = 
+      import quotes.reflect.* 
+      
+      // repeated
+      def isObjectOrScala(bc: Symbol) =
+        bc.name.contains("java.lang.Object") || bc.fullName.startsWith("scala.")
+
+      def fromConstructor(from: Symbol): List[(String, List[Term])] =
```
adamw: shouldn't this be `annotationsFromConstructor`?

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+          .groupBy { case (name, _) => name }
+          .toList
+          .map { case (name, l) => name -> l.flatMap(_._2) }
+
+      def selectTermsByName(tss: List[(String, List[Term])], name: String): List[Term] = 
+        tss
+          .toMap
+          .getOrElse(name, Nil)
+      
+      def filterAnnotation(annTerm: Term): Boolean =
```
adamw: this looks the same as a private method above, cannot we share it?

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+        vCtor.paramSymss match 
+          case List(paramSymbol: Symbol) :: Nil => 
+            val paramTypeTree = 
+              paramSymbol.tree match 
+                case v: ValDef => v.tpt
+                case _ => report.errorAndAbort("Error handling param symbol tree '${paramSymbol.tree}'")
+            val aTpe = paramTypeTree.tpe
+            argApp.select(as).appliedToType(aTpe)
+
+          case _ => report.errorAndAbort("Error handling param symbol tree '${paramSymbol.tree}'")
```
adamw: there's the `s` interpolator missing

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+      val argApp = argsTerm.select(appl).appliedTo(Literal(IntConstant(0)))
+      val as = argApp.tpe.typeSymbol.methodMember("asInstanceOf").head
+
+      val arg = 
+        vCtor.paramSymss match 
+          case List(paramSymbol: Symbol) :: Nil => 
+            val paramTypeTree = 
+              paramSymbol.tree match 
+                case v: ValDef => v.tpt
+                case _ => report.errorAndAbort("Error handling param symbol tree '${paramSymbol.tree}'")
```
adamw: same here - no `s`

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+
+          val app = 
+            Apply(
+              fun = applyFuncTerm, 
+              args = args
+              )
+
+          app.asExprOf[CaseClass.Param[TC, A]]
+        
+        case _ => report.errorAndAbort("Error handling symbol '${sym.name}'")
```
adamw: and here, and above

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+              case _ => report.errorAndAbort("Error handling param symbol tree '${paramSymbol.tree}'")
+          
+          val paramTypeTpe = paramTypeTree.tpe
+          val paramTC = TypeRepr.of[TC].appliedTo(paramTypeTpe)
+
+          val paramCallByNeed = paramTC.asType match 
+            case '[t] => 
+              Expr
+                .summon[t]
+                .map{ e => e.asCallByNeedExpr }
```
adamw: is this formatted with scalafmt?

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+        def asCallByNeed(tpe: TypeRepr): Term = 
+          Select(
+            '{CallByNeed}.asTerm, 
+            TypeRepr.of[CallByNeed.type].termSymbol.declaredMethod("apply").head
+            )
+            .appliedToType(tpe)
+            .appliedTo(term)
+
+      extension (tpe: TypeRepr)
+        def defaultValues: Map[String, Expr[Any]] =
```
adamw: maybe we can extract some common code with `defaultValue[T]` above?

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+          }
+          .filter(_._2.nonEmpty)
+
+      def getDefaultValue[P: Type](dict: Map[String, Expr[Any]], name: String): Expr[CallByNeed[Option[Any]]] = 
+        dict.get(name) match 
+          case Some(e) => '{ new CallByNeed(() => Some( ($e).asInstanceOf[P] )) }
+          case None => '{ new CallByNeed(() => None) }
+
+      extension (term: Term)
+        def asCallByNeed(tpe: TypeRepr): Term =
```
adamw: is this used?

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+          
+
+      val aTpe: TypeRepr = TypeRepr.of[A]
+      val aSym: Symbol = aTpe.typeSymbol
+      val aCtor: Symbol = aSym.primaryConstructor
+
+      val defaultValuesDict = aTpe.defaultValues
+      val annsDict = getAnnsDict(aTpe)
+      val inheritedAnnsDict = getInheritedAnnsDict(aTpe)
+      val typeAnnsDict = getTypeAnnsDict(aTpe)
```
adamw: why not convert to a map once, instead of on every usage? if it's a dict

src/core/macro.scala
```diff
@@ -259,12 +281,223 @@ object Macro:
...
+
+      def getAnnsDict(typeRepr: TypeRepr): List[(String, List[Term])] = 
+        val annTerms: List[(String, List[Term])] = 
+          (fromConstructor(typeRepr.typeSymbol) ++ fromDeclarations(typeRepr.typeSymbol))
+            .filter {case (_, ts) => ts.nonEmpty}
+        groupTermsByName(annTerms)
+
+      def getInheritedAnnsDict(typeRepr: TypeRepr): List[(String, List[Term])] = 
+        val annTerms: List[(String, List[Term])] = 
+          typeRepr.baseClasses
```
adamw: this also looks very simlar if not identical to `inheritedAnns`; maybe we can call this method from the one accepting a `Type`?

adamw (changes requested):
there's some code duplication which I think we can remove

src/core/magnolia.scala
micsza: Yes, made them private.

src/core/magnolia.scala
micsza: Refactored.

src/core/impl.scala
micsza: :) fixed.

build.sbt
micsza: Removed.

src/test/ValueClassesTests.scala
micsza: Added.

src/test/SumsTests.scala
micsza: refactored.

src/core/macro.scala
micsza: I have nested a somehow specific `ValueClassDerivation` object inside a somehow more general `Macro` object. Macro implementations for VC have been made private. The already existing one cannot be marked private without compatibility issues though.

src/core/macro.scala
micsza: No longer used but I have renamed the very same method in `CollectAnnotations` class.

src/core/macro.scala
micsza: Indeed - it was due to  Quotes being path-dependent and I had some troubles with accessing that. Now fixed and shared .

src/core/macro.scala
micsza: Fixed.

src/core/macro.scala
micsza: Fixed.

src/core/macro.scala
micsza: Fixed.

src/core/macro.scala
micsza: Refactored.

src/core/macro.scala
micsza: Refactored.

src/core/macro.scala
micsza: Hmm, apparently it was .. Anyway, I have manually simplified to `_.m`.

src/core/macro.scala
micsza: Refactored.

src/core/macro.scala
micsza: Ooops, nope. Removed.

src/core/interface.scala
micsza: Still stuck with it. I have issues of mapping from `Expr[List[Any]]` to `Expr[IArray[Any]]`.

micsza: This extra constructor is now removed. Many thanks for help!

src/test/ValueClassesTests.scala
adamw: Thanks - but what's a product value class? Do you mean case classes? If so, `ServiceName1` is a case class, but it is disabled?

src/test/SumsTests.scala
```diff
@@ -101,10 +101,7 @@ class SumsTests extends munit.FunSuite:
   ) {
     val error = compileErrors("Show.derived[Parent]")
     assert(
-      error contains "No given instance of type deriving.Mirror.Of[magnolia1.tests.SumsTests.Parent] was found for parameter x$1 of method derived in trait Derivation."
-    )
-    assert(
-      error contains "trait Parent is not a generic sum because its child trait BadChild is not a generic product because it is not a case class"
+      error contains "Deriving the typeclass based on mirrors or directly is not possible for \"magnolia1.tests.SumsTests.Parent\":String. Please refer to the documentation or report a feature request."
```
adamw: what's this `:String`?

src/core/macro.scala
```diff
@@ -192,79 +211,323 @@ object Macro:
...
+        symbol.toList
+          .map(_.tree)
+          .flatMap {
+            case ClassDef(_, _, parents, _, _) =>
+              parents
+                .collect {
+
+                  case t: TypeTree => t.tpe
+
+                  // case for AnyVal type annotations: with "-Yretain-trees" scalac option, the TypeTree of the annotation gets erased,
```
adamw, line 228: with the option, or without the option? It would seem on first sight that trees get erased without `retain-trees` :)

adamw (approved):
Looks good, thanks! :)

src/test/ValueClassesTests.scala
micsza: Yes, I mean case classes - `ServiceName` test is now enabled.

src/test/SumsTests.scala
micsza: Improved.

src/core/macro.scala
micsza: Tricky as it seems, it is **with** that option. It appears that `-Yretain-trees` may change the types of the trees, f.e. https://github.com/lampepfl/dotty/issues/14195.

PR #441, opened 2022-12-22
Revert "Support for value classes"

micsza:
Reverts softwaremill/magnolia#435 due to binary compatibilty issues.


eejbyfeldt:
What would be required to reintroduce this change?

adamw:
@eejbyfeldt we'll get there ;) there's a `v2` branch with those changes, unfortunately it diverged from the main scala3 branch. See: https://github.com/softwaremill/magnolia/issues/552

PR #457, opened 2023-02-06
Add isSingletonCasesEnum flag

sake92:
It is useful to disambiguate between "singleton cases"/simple enums and ADTs.

My concrete use case was JSON serialization:
- simple enums get written as simple strings
- ADTs get written as objects with `@type`  field, and other fields if they have them

Implementation-wise, the [official docs](https://docs.scala-lang.org/scala3/reference/enums/desugarEnums.html#translation-of-enums-with-singleton-cases) say that only "singleton cases" enums will have a `values` method.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import com.softwaremill.UpdateVersionInDocs
 import com.softwaremill.SbtSoftwareMillCommon.commonSmlBuildSettings
 import com.softwaremill.Publish.{updateDocs, ossPublishSettings}
 
-val scala3 = "3.2.1"
+val scala3 = "3.2.2"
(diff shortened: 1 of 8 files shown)


adamw:
Thanks - this isn't binary-compatible, but there's a `scala3-v2` branch that can accept these changes :)

sake92:
Thanks @adamw !   
Do those artifacts get published too?

adamw:
Not yet - but we'll get there eventually :)

sake92:
Could this be merged now?

adamw:
@sake92 see https://github.com/softwaremill/magnolia/issues/552

PR #461, opened 2023-03-02
Consistent ordering for sealed traits

note:
I stumbled upon a difference in order of `SealedTrait.subtypes` between (magnolia1_3 % 1.2.7) and (magnolia1_2 % "1.1.3") when working on some project I intend to publish for both scala 2 and 3

It mimics the logic introduced in the original magnolia in this PR: https://github.com/softwaremill/magnolia/pull/196 and still [present in magnolia1_2 at 1.1.3](https://github.com/softwaremill/magnolia/blob/a4eb9bd189d9fcb36eabd01ff2eefb137ff26ca7/core/src/main/scala/magnolia1/magnolia.scala#L617)

### DIFF
--- a/src/test/SumsTests.scala
+++ b/src/test/SumsTests.scala
@@ -4,6 +4,76 @@ import magnolia1.*
 import magnolia1.examples.*
 import scala.annotation.StaticAnnotation
 
+object SumsTests:
+
+  case class MyAnnotation(order: Int) extends StaticAnnotation
+  case class MyTypeAnnotation(order: Int) extends StaticAnnotation
+
+  sealed trait Entity
+  case class Company(name: String) extends Entity
+  case class Person(name: String, age: Int) extends Entity
+  case class Address(line1: String, occupant: Person)
+
+  sealed trait Color
+  case object Red extends Color
+  case object Green extends Color
+  case object Blue extends Color
+  case object Orange extends Color
+  case object Pink extends Color
+
+  sealed trait Y
+  case object A extends Y
+  case class B(s: String) extends Y
+
+  enum Size:
+    case S, M, L
+
+
+  sealed trait Sport
+  case object Boxing extends Sport
+  case class Soccer(players: Int) extends Sport
+
+  sealed trait Complex
+  object Complex:
+    case object Object extends G
+    sealed trait A extends Complex
+    sealed trait B extends A
+    case object ObjectC extends Complex
+    case object ObjectD extends A
+    case object ObjectE extends B
+    case object ObjectF extends A with Complex
+    sealed trait G extends B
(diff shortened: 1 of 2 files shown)


src/test/SumsTests.scala
```diff
@@ -4,6 +4,76 @@ import magnolia1.*
 import magnolia1.examples.*
 import scala.annotation.StaticAnnotation
 
+object SumsTests:
```
note, line 7: I moved the whole `SumsTests` to reenable `derive all subtypes in complex hierarchy` test. Yes, it's [a compiler bug](https://github.com/lampepfl/dotty/issues/17034) but such a seemingly irrelevant change makes it go away.

src/test/SumsTests.scala
```diff
@@ -131,97 +201,24 @@ class SumsTests extends munit.FunSuite:
...
-  //   ).foreach { o =>
-  //     val chosen = res.choose(o)(identity)
-  //     assertEquals(chosen.value, o)
-  //     assertEquals(
-  //       chosen.typeInfo.short,
-  //       o.getClass.getSimpleName.replace("$", "")
-  //     )
-  //   }
-  // }
+   test("derive all subtypes in complex hierarchy".only) {
```
note: Oops, forgotten removing `only`, pushing a change

src/test/SumsTests.scala
```diff
@@ -131,97 +201,24 @@ class SumsTests extends munit.FunSuite:
...
+       s"$pkg.ClassH",
+       s"$pkg.Object",
+       s"$pkg.ObjectC",
+       s"$pkg.ObjectD",
+       s"$pkg.ObjectE",
+       s"$pkg.ObjectF",
+       s"$pkg.Scoped.Object",
+     )
+
+     assertEquals(res.subtypes.map(_.typeInfo.full).toList, expected)
```
note, line 218: Yes, I changed test methodology a bit here, but the previous one was problematic as

```scala
     List(
       Complex.ObjectE,
       Complex.ObjectC,
       Complex.ObjectE,
       Complex.ObjectE,
       Complex.ObjectE,
       Complex.ObjectE,
       Complex.ObjectE
     ).foreach { o =>
       val chosen = res.choose(o)(identity)
       assertEquals(chosen.value, o)
       assertEquals(
         chosen.typeInfo.short,
         o.getClass.getSimpleName.replace("$", "")
       )
     }
```

passes tests just fine

src/test/SumsTests.scala
adamw: * compiler feature ;)

Sure thing, looks good

adamw:
Thank you!

PR #465, opened 2023-05-05
bugfixes for scala3

katlasik:
- Fix bug with getting default value for parameters - cast exception is no longer thrown if type is different
- Fix OtherTests


adamw (approved):
It would be great to separate the reformatting & logic changes into separate PRs (next time :) )

PR #468, opened 2023-05-24
Dealias types when deriving names (Scala 3)

### LINKED ISSUE #467: Type parameter names should should be dealiased for Scala 3
This is a follow-up to https://github.com/softwaremill/tapir/issues/2911

The issue in tapir is caused by Magnolia derivation, which doesn't `.dealias` type parameter names.

This issue doesn't happen for Scala 2, where dealiasing seems to be correctly applied.

kciesielski:
Fixes #467 for Scala 3

This PR ensures that type names are derived from dealiased underlying types.
Inspired by an issue reported in tapir:

```scala
//> using scala 3.3.0-RC6
//> using dep com.softwaremill.sttp.tapir::tapir-core:1.4.0
import sttp.tapir.*
import sttp.tapir.generic.auto.*

case class Foo[T](t: T)

object Domain1:
  type Type = Int

object Domain2:
  type Type = String

@main
def main =
  val d1Schema = summon[Schema[Foo[Domain1.Type]]]
  val d2Schema = summon[Schema[Foo[Domain2.Type]]]
  println(d1Schema.name.map(_.show)) // Some(.Foo[Type]), should be Some(.Foo[Int])
  println(d2Schema.name.map(_.show)) // Some(.Foo[Type]), should be Some(.Foo[String])
```

This issue doesn't happen for Scala 2, where dealiasing seems to be correctly applied.

### DIFF
--- a/src/core/macro.scala
+++ b/src/core/macro.scala
@@ -137,11 +137,11 @@ object Macro:
 
     def normalizedName(s: Symbol): String =
       if s.flags.is(Flags.Module) then s.name.stripSuffix("$") else s.name
-    def name(tpe: TypeRepr): Expr[String] = tpe match
-      case TermRef(typeRepr, name) if tpe.typeSymbol.flags.is(Flags.Module) =>
+    def name(tpe: TypeRepr): Expr[String] = tpe.dealias match
+      case matchedTpe @ TermRef(typeRepr, name) if matchedTpe.typeSymbol.flags.is(Flags.Module) =>
         Expr(name.stripSuffix("$"))
       case TermRef(typeRepr, name) => Expr(name)
-      case _                       => Expr(normalizedName(tpe.typeSymbol))
+      case matchedTpe              => Expr(normalizedName(matchedTpe.typeSymbol))
 
     def ownerNameChain(sym: Symbol): List[String] =
       if sym.isNoSymbol then List.empty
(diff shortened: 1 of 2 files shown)


src/core/macro.scala
```diff
@@ -164,7 +164,7 @@ object Macro:
           )
         }
       case _ =>
-        '{ TypeInfo(${ owner(tpe) }, ${ name(tpe) }, Nil) }
+        '{ TypeInfo(${ owner(tpe) }, ${ name(tpe.dealias) }, Nil) }
```
KacperFKorban: I think that we should also `dealias` when computing the owner. Otherwise, the full name will be very confusing. For one of the examples it would be `magnolia1.tests.ProductsTests.Domain1.Int`.

Also the same dealiasing should be done in the `AppliedType` case, so maybe it would be easier to just move it into `name` and `owner` methods. It should be safe enough. i.e.
```scala
  def name(tpe: TypeRepr): Expr[String] = tpe.dealias match
    case tpe@TermRef(typeRepr, name) if tpe.typeSymbol.flags.is(Flags.Module) =>
      Expr(name.stripSuffix("$"))
    case TermRef(typeRepr, name) => Expr(name)
    case tpe                       => Expr(normalizedName(tpe.typeSymbol))
...
  def owner(tpe: TypeRepr): Expr[String] = Expr(
    ownerNameChain(tpe.dealias.typeSymbol.maybeOwner).mkString(".")
  )
```

kciesielski: Thanks @KacperFKorban, that would be indeed much better, I updated the PR.

KacperFKorban (approved):

PR #481, opened 2023-07-13
Reduce amount of inlined code, this increases compilation speed by 50% and brings it back to scala2.13 level

OlegYch:


src/core/impl.scala
```diff
@@ -24,47 +24,66 @@ object CaseClassDerivation:
         defaultValue[A].to(Map)
       )*
     )
-
-    new CaseClass(
+    ProductCaseClass(
```
adamw, line 27: sounds awesome! Out of curiosity, can you share some insight on *why* this improves performance? maybe even as a comment, for informing future reads and avoiding introducing regressions

I'm guessing that we're not generating a new anonymous class construction every time, but instead we generate an invocation of the constructor, right? but why does this have such a performance impact?

OlegYch: i think it's just the amount of inlined code, compiler has to compile it for every method invocation
but i'm not really sure...

joroKr21: Well you need a separate classfile for every anonymous instance so it can add up.

adamw: thanks, I guess the simplest explanation is the the right one ;)

PR #485, opened 2023-08-14
Split subtypes array initialization to multiple functions.

nox213:
We are using Magnolia in production and mainly use it to derive the parent type's type class instances from the type class instances of subtypes. One of our types has more than 700 subtypes, and when derives typeclass instances of this type, the process of initializing the huge subtypes array is performed in a single method, resulting in a "Method too large" compilation error. This PR splits the initialization of the subtypes array into multiple functions and calls each function at the end to combine the array. This method also increases the size of the method, but to a much slower degree, so I think it's still worthwhile. I would appreciate it if you could review it.

### DIFF
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -673,12 +673,12 @@ object Magnolia {
           .unzip
 
         val subtypesVal = c.freshName(TermName("subtypes"))
-        val combineArrays = functionNames.map(name => q"""$subtypesVal ++= $name""")
+        val combinations = functionNames.map(name => q"""$subtypesVal ++= $name""")
 
         Some(q"""{
           ..$partialAssignmentFunctions
           val $subtypesVal = Array.newBuilder[$subType]
-          ..$combineArrays
+          ..$combinations
           $typeNameDef
           ${c.prefix}.split(new $SealedTraitSym(
             $typeName,
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -629,7 +629,6 @@ object Magnolia {
           error(s"could not find any direct subtypes of $typeSymbol")
         }
 
-        val subtypesVal = c.freshName(TermName("subtypes"))
         val typeclasses = for (subType <- subtypes) yield {
           val path = CoproductType(genericType.toString)
           val frame = stack.Frame(path, resultType, assignedName)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -640,29 +639,50 @@ object Magnolia {
...
+          .unzip
+
+        val subtypesVal = c.freshName(TermName("subtypes"))
+        val combinations = functionNames.map(name => q"""$subtypesVal ++= $name""")
+
         Some(q"""{
-          val $subtypesVal = new $ArrayClass[$subType](${assignments.size})
-          ..$assignments
+          ..$partialAssignmentFunctions
+          val $subtypesVal = Array.newBuilder[$subType]
```
joroKr21: Pass the array and index to the helper methods, no need for a builder

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -640,29 +639,50 @@ object Magnolia {
...
+              val (annotationTrees, inheritedAnnotationTrees) = annotationsOf(symbol)
+              q"""$partialSubtypesVal($idx) = $SubtypeObj[$typeConstructor, $genericType, $subType](
+                        ${typeNameOf(subType)},
+                        $idx,
+                        $ArrayObj(..${annotationTrees}): Array[_root_.scala.Any],
+                        $ArrayObj(..${inheritedAnnotationTrees}): Array[_root_.scala.Any],
+                        $ArrayObj(..${typeAnnotationsOf(symbol, fromParents = true)}): Array[_root_.scala.Any],
+                        $CallByNeedObj($typeclass),
+                        (t: $genericType) => t.isInstanceOf[$subType],
+                        (t: $genericType) => t.asInstanceOf[$subType]
```
joroKr21: You can extract those as helper methods in the beginning to reduce the method size

core/src/main/scala/magnolia1/magnolia.scala
nox213: [4acf714](https://github.com/softwaremill/magnolia/pull/485/commits/4acf7140732b01d41f7ec61c9ad2bc639c260704)

nox213:
@joroKr21 I fixed based on what you commented. Could you review this PR again?

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -393,6 +393,31 @@ object Magnolia {
           $impl
       """
 
+      def constructPartialAssignmentFunction(typeclasses: List[(Type, Tree)], arrayType: Type): (TermName, Tree) = {
+        val name = c.freshName(TermName("partialAssignments"))
+        val assignments = typeclasses.zipWithIndex.map { case ((subType, typeclass), idx) =>
+          val symbol = subType.typeSymbol
+          val (annotationTrees, inheritedAnnotationTrees) = annotationsOf(symbol)
+          q"""arr(start + $idx) = $SubtypeObj[$typeConstructor, $genericType, $subType](
```
joroKr21: `arr` and `start` need to be fresh names for hygiene.

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -393,6 +393,31 @@ object Magnolia {
...
+            ${typeNameOf(subType)},
+            $idx,
+            $ArrayObj(..${annotationTrees}): Array[_root_.scala.Any],
+            $ArrayObj(..${inheritedAnnotationTrees}): Array[_root_.scala.Any],
+            $ArrayObj(..${typeAnnotationsOf(symbol, fromParents = true)}): Array[_root_.scala.Any],
```
joroKr21: Again for hygiene:
```suggestion
            $ArrayObj[$AnyTpe](..${annotationTrees}),
            $ArrayObj[$AnyTpe](..${inheritedAnnotationTrees}),
            $ArrayObj[$AnyTpe](..${typeAnnotationsOf(symbol, fromParents = true)}),
```

core/src/main/scala/magnolia1/magnolia.scala
joroKr21: I guess it's not so easy, because if we make helper methods they need to deal with `ClassTag` 🤔

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -393,6 +393,31 @@ object Magnolia {
...
+            $CallByNeedObj($typeclass),
+            (t: $genericType) => t.isInstanceOf[$subType],
+            (t: $genericType) => t.asInstanceOf[$subType]
+          )"""
+        }
+        val tree =
+          q"""def $name(arr: Array[$arrayType], start: Int) = {
```
joroKr21: Extract this to the top level:
```scala
val ArrayTpe = typeOf[Array[Any]].typeConstructor
```

And use it like this:
```suggestion
          q"""def $name(arr: $ArrayTpe[$arrayType], start: $IntTpe) = {
```

core/src/main/scala/magnolia1/magnolia.scala
nox213: Ah, I understand what you mean. Yes, it seems we need to generate sort of runtime reflection codes by macro if we want extract that codes as helper method. So I would say keeping the current way is better even if the runtime reflection code would reduce the method size dramatically. What do you think?

core/src/main/scala/magnolia1/magnolia.scala
nox213: [be8bd49](https://github.com/softwaremill/magnolia/pull/485/commits/be8bd496fe5891d6f7fea62ed31e3a3122cf2b8c)

core/src/main/scala/magnolia1/magnolia.scala
nox213: [be8bd49](https://github.com/softwaremill/magnolia/pull/485/commits/be8bd496fe5891d6f7fea62ed31e3a3122cf2b8c)

core/src/main/scala/magnolia1/magnolia.scala
nox213: [be8bd49](https://github.com/softwaremill/magnolia/pull/485/commits/be8bd496fe5891d6f7fea62ed31e3a3122cf2b8c)
I fixed it. Could you explain me the reason why we need fresh names even if they are local variables of that method?

core/src/main/scala/magnolia1/magnolia.scala
joroKr21: Yes, let's keep it like this

core/src/main/scala/magnolia1/magnolia.scala
joroKr21: Because the user code might contain the same names. It's just insurance.

joroKr21 (approved):
LGTM 👍

adamw:
Thanks! Release is in progress

Issue #487, opened 2023-08-28
magnolia scala2-v1.1.5 made generated type-classes not serializable


RustedBones:
Last release of magnolia `scala2-v1.1.5` introduced a regression, generating not serializable type-classes:

> java.io.NotSerializableException: magnolify.cats.HashDerivationSuite$PartialAssignments$macro$111$1$

See test failures in scala-steward [PR](https://github.com/spotify/magnolify/pull/823)

adamw:
Thanks for the report - can you just double-check that 1.1.4 works fine?

RustedBones:
We've already updated to 1.1.4 in main. The [PR](https://github.com/spotify/magnolify/pull/808) passed on the same tests.
I think this is the addition of the new object in a non serializable class that causes the issue as object outer must be serializable.

adamw:
Releasing 1.1.6 with the problematic PR reverted

PR #489, opened 2023-08-29
Revert "Reduce constant pool usage, this make Magnolia able to derive instance for sealed trait that has a lot of subtypes "

adamw:
Due to https://github.com/softwaremill/magnolia/issues/487


joroKr21:
I guess just make the object serializable

PR #495, opened 2023-10-21
Add filter for `jdk.internal` annotations

### LINKED ISSUE #493: [Scala 3] `Macro.anns` throws `IllegalAccessError` when used on parameters with Java types on JDK17+
Hi there 👋

There seems to be an issue when using `Macro.anns` on case class parameters that are Java types on JDK17+ (works fine on JDK11). I created a small reproduction via scatsie [here](https://scastie.scala-lang.org/9dtxN8fiRzGqZH1XONiOjg).

The fix to the issue is quite straightforward, and it's to add the following condition to the `filterAnnotation` method: `a.tpe.typeSymbol.owner.fullName != "jdk.internal"`. Please let me know if you'd like me to create a PR for it and I'll do so

Please note that AFAICT, `Macro.paramAnns` works as intended, although I thought to let you know since the fix is relatively simple and there might be edge cases where a user will want to extract the param annotations manually as in the repro above

kyri-petrou:
Fixes #493

I haven't added a test for this case since the CI is run on JDK11 only. Do you prefer if I add a test as in the issue repro in the issue above?

### DIFF
--- a/src/core/macro.scala
+++ b/src/core/macro.scala
@@ -266,6 +266,6 @@ object Macro:
 
     private def filterAnnotation(a: Term): Boolean =
       a.tpe.typeSymbol.maybeOwner.isNoSymbol ||
-        a.tpe.typeSymbol.owner.fullName != "scala.annotation.internal" ||
-        a.tpe.typeSymbol.owner.fullName != "jdk.internal"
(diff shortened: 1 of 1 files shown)


adamw:
It seems there are some test failures

kyri-petrou:
Oops, seems to have used `||` instead of `&&` 🤦 It should be good now

PR #498, opened 2023-11-13
[scala2] Add test for serializable generated type-class

RustedBones:
Adding a unit-test to avoid regression like https://github.com/softwaremill/magnolia/issues/487


adamw:
Thank you! Great to expand the test suite :)

PR #499, opened 2023-11-13
[scala3] Add test for serializable generated type-class

RustedBones:
Scala3 version of magnolia does not generate serializable typeclass when defined inside a non-serializable outer, unlike the scala 2 version. Tested in https://github.com/softwaremill/magnolia/pull/498.

It looks like the `summonInline[Typeclass[p]]` lambda in the `CallByNeed` takes a reference on the outer obejct.


RustedBones:
@adamw since you reviewed https://github.com/softwaremill/magnolia/pull/498

This new test is only relevant for JVM, and fails for JS and native.
The project layout for magnolia looks non standard. I don't know how to add JVM only test.

Would migrating to `sbt-corssproject` be considered ? I can submit a PR for that.

adamw:
@RustedBones our projects use https://github.com/sbt/sbt-projectmatrix, in my experience this is the only sane way to maintain projects with both simple & complex cross-platform setups. I think adding the test in `scalajvm` would do the trick (as opposed to `scala`), see e.g. [here](https://github.com/softwaremill/tapir/blob/bdcc324b407199673d17aebc9e7cae95d80baec7/core/src/test/scalajvm/sttp/tapir/CodecTestDateTime.scala#L42).

RustedBones:
Ok, let's stick with `sbt-projectmatrix`.
At the moment, there isn't a `scala` folder. The project defines fake project root and overrides the `scalaSource`. (btw, this setup does not work well with intelliJ and requires to manually add source to get indexer working).
What about standardizing the project layout like [this](https://github.com/softwaremill/magnolia/compare/scala3...RustedBones:magnolia:project-structure?expand=1) and create a `scalajvm` source folder in the test project ?

RustedBones:
Any take on this one?
There is probably a trick to avoid this but I couldn't find any other alternative

turb:
It would be great if this PR can be reviewed / merged, since it's blocking scio migration towards Scala 3.

adamw:
@turb ah sorry, this somehow fell through the cracks. There's a conflict, otherwise it looks good.

adamw:
@turb would you be able to create an updated PR maybe?

turb:
> @turb would you be able to create an updated PR maybe?

I suppose @RustedBones may prefer to do it himself since it's his PR, but if he's busy I will.

core/src/main/scala/magnolia1/impl.scala
```diff
@@ -97,12 +105,17 @@ object CaseClassDerivation:
...
+
+        val d = new SerializableFunction0[Option[p]]:
+          private def unsafeCast(any: Any) = Option.when(any == null || (any: @unchecked).isInstanceOf[p])(any.asInstanceOf[p])
+          override def apply(): Option[p] = defaults.get(label).flatten.flatMap(d => unsafeCast(d.apply))
         paramFromMaps[Typeclass, A, p](
           label,
-          CallByNeed(summonInline[Typeclass[p]]),
-          CallByNeed.withValueEvaluator(defaults.get(label).flatten.flatMap(d => unsafeCast(d.apply))),
+          new CallByNeed(tc),
+          new CallByNeed(d, () => true),
```
adamw: I think this makes `.withValueEvaluator` unused? Maybe we could move the comment here, as what's the purpose of the `() => true`?

RustedBones: Added `by-value` API for clarity

adamw:
Thanks all! :)

Issue #258, opened 2020-05-26
Support the way to use only `dispatch` except `combine` safely


taeguk:
This is small example for showing what I want to say:
```scala
import language.experimental.macros
import magnolia._
import cats.Show

object ShowDerivation {
  type Typeclass[T] = Show[T]

  def combine[T](ctx: CaseClass[Show, T]): Show[T] = new Show[T] {
    def show(value: T): String =
      throw new Exception("Don't support combine for show derivation.")
  }

  def dispatch[T](ctx: SealedTrait[Show, T]): Show[T] = new Show[T] {
    def show(value: T): String =
      ctx.dispatch(value)(sub => sub.typeclass.show(sub.cast(value)))
  }

  def gen[T]: Show[T] = macro Magnolia.gen[T]
}

sealed trait MySealedTrait
case class MyCaseClass(a: Int) extends MySealedTrait
case object MyCaseObject       extends MySealedTrait

// There is the compile error:
// magnolia: could not find Show.Typeclass for type Int
//     in parameter 'a' of product type ForFun.MyCaseClass
val a = ShowDerivation.gen[MyCaseClass]

// There is no compile error.
// I want to make a compilation fails here.
val b = ShowDerivation.gen[MyCaseObject.type]
```

Currently, I use `magnolia` for an automatic materialization of `sealed trait` when materializations of derived types (`case class` and `case object`) are given manually.
In this usage of `magnolia`, I only need `dispatch` except `combine`. And I want to make a compilation fails if there is no materialization of any derived type to prevent mistakes.

I think it can be helpful in many cases to support the way to use only `dispatch` except `combine` safely.

PR #501, opened 2023-12-12
Enrich error message when derivation fails.

nox213:
We're using this pattern(#258) in our project quiet often, and I think it would be great if it provides better error message.
For example,
```
package magnolia1.examples

import magnolia1.{Magnolia, ReadOnlyCaseClass, SealedTrait}

trait Print[T] {
  def print(t: T): String
}

trait GenericPrint {
  type Typeclass[T] = Print[T]

  // we only have split method
  def split[T](ctx: SealedTrait[Print, T]): Print[T] = { value =>
    ctx.split(value) { sub =>
      sub.typeclass.print(sub.cast(value))
    }
  }

  implicit def gen[T]: Print[T] = macro Magnolia.gen[T]
}

sealed trait Entity
case class Person(name: String, age: Int) extends Entity
case class Place(name: String) extends Entity

object Print extends GenericPrint {
  implicit val person: Print[Person] = p => p.name + " is " + p.age
//  implicit val place: Print[Place] = _.name
  implicit val entity: Print[Entity] = gen
}

object PrintExample extends App {
  println(implicitly[Print[Entity]].print(Person("John", 32)))
  println(implicitly[Print[Entity]].print(Place("London")))
}
```
If we don't have an instance for `Place`, the current error message is
`the method 'join' must be defined on the derivation object Print to derive typeclasses for case classes`
So, it tells you that it needs `join` method to derive a type class instance, but it doesn't tell you which instance it was trying to derive.

With this PR, it prints,
`magnolia: unable to derive magnolia1.examples.Print.Typeclass[magnolia1.examples.Place] -- the method 'join' must be defined on the derivation object Print to derive typeclasses for case classes`

This one tells which one it was trying to derive.

### DIFF
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -258,12 +258,16 @@ object Magnolia {
 
     def extractParameterBlockFor(termName: String, category: String): List[Symbol] = {
       val term = TermName(termName)
-      val searchType = stack.top.map(_.searchType.toString).getOrElse("")
+      val searchType = stack.top.map(_.searchType.toString)
+      val additionalInfo = searchType match {
+        case Some(tpe) => s"unable to derive $tpe -- "
+        case None      => ""
+      }
       val classWithTerm = c.prefix.tree.tpe.baseClasses
         .find(cls => cls.asType.toType.decl(term) != NoSymbol)
         .getOrElse(
           error(
-            s"unable to derive $searchType -- the method `$termName` must be defined on the derivation $prefixObject to derive typeclasses for $category"
+            s"${additionalInfo}the method `$termName` must be defined on the derivation $prefixObject to derive typeclasses for $category"
           )
         )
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -258,9 +258,18 @@ object Magnolia {
 
     def extractParameterBlockFor(termName: String, category: String): List[Symbol] = {
       val term = TermName(termName)
+      val searchType = stack.top.map(_.searchType.toString)
+      val additionalInfo = searchType match {
+        case Some(tpe) => s"unable to derive $tpe -- "
+        case None      => ""
+      }
```
joroKr21: This should move inside `getOrElse` to avoid doing work on the happy path.

nox213: [e036de4](https://github.com/softwaremill/magnolia/pull/501/commits/e036de4a74e7ad2c829bb99e4131a54de4490dc0)
Thanks for the review.

joroKr21 (approved):
👍

adamw:
Thanks!

PR #505, opened 2024-01-09
[scala3] Use standard project structure

RustedBones:
Use a standard project structure. Would help with #499, to add a `test/src/test/scalajvm` test source folder for `test3` project only

From
```
src/
  core/
  examples/
  test/
  test-enum/ <-- was no referenced by build
```
to
```
core/src/main/scala/magnolify1/
examples/src/main/scala/magnolify1/examples/
test/src/test/
  java/magnolify1/tests/
  scala/magnolify1/tests/
```


adamw:
Thanks!

PR #533, opened 2024-05-20
[scala2] feat: adding on-demand default evaluation

Andrapyre:
## Description
Currently, all default values are precompiled. This has led to random generators, such as random `UUID.randomUUID()` and `Instant.now()` being memoized (see https://github.com/zio/zio-json/issues/1055). This PR adds an option to evaluate default parameters on-demand. See the accompanying PR for Scala 3 support: https://github.com/softwaremill/magnolia/pull/534

### DIFF
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -963,6 +963,9 @@ object CallByNeed {
 // The supportDynamicValueEvaluation is passed as a function so that it can be nullified. Otherwise, there is no need for the function value.
 final class CallByNeed[+A] private (private[this] var eval: () => A, private var supportDynamicValueEvaluation: () => Boolean)
     extends Serializable {
+
+  def this(eval: () => A) = this(eval, () => false)
+
   val valueEvaluator: Option[() => A] = {
     val finalRes = if (supportDynamicValueEvaluation()) {
       val res = Some(eval.fv)
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -5,6 +5,7 @@ import magnolia1.Monadic.Ops
 import scala.annotation.{compileTimeOnly, tailrec}
 import scala.collection.mutable
 import scala.language.higherKinds
+import scala.reflect.ClassTag
 import scala.reflect.macros._
 
 /** the object which defines the Magnolia macro */
@@ -946,6 +947,7 @@ private[magnolia1] object CompileTimeState {
 
 object CallByNeed { def apply[A](a: => A): CallByNeed[A] = new CallByNeed(() => a) }
 final class CallByNeed[+A](private[this] var eval: () => A) extends Serializable {
+  val uncalledValue: () => A = eval.fv
   lazy val value: A = {
     val result = eval()
     eval = null
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -5,7 +5,6 @@ import magnolia1.Monadic.Ops
 import scala.annotation.{compileTimeOnly, tailrec}
 import scala.collection.mutable
 import scala.language.higherKinds
-import scala.reflect.ClassTag
 import scala.reflect.macros._
 
 /** the object which defines the Magnolia macro */
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -946,19 +946,26 @@ private[magnolia1] object CompileTimeState {
 }
 
 object CallByNeed {
-  def apply[A](a: => A): CallByNeed[A] = new CallByNeed(() => a, false)
-  def withValueEvaluator[A](a: => A): CallByNeed[A] = new CallByNeed(() => a, true)
+  def apply[A](a: => A): CallByNeed[A] = new CallByNeed(() => a, () => false)
+  def withValueEvaluator[A](a: => A): CallByNeed[A] = new CallByNeed(() => a, () => true)
 }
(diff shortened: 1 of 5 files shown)


adamw:
Thanks for the PR! I'll take a detailed look in a couple of days, but for sure we'll need to find a way to keep bincompat. Or move these changes to v2.

joroKr21:
> Since this would break new CallByNeed(param) and force users to use CallByNeed(param), I tried to find some examples of libs where users would need to make this change, but I couldn't find any instances outside of Magnolia where this implementation is actually being used. See [this search](https://github.com/search?q=%22new+CallByNeed%22&type=code&p=1). It seems that breaking this would likely have minimal effect, if any.

It can break when you have two dependencies that use Magnolia on the classpath, one the older version using the constructor, and another one the newer version using `apply`. You can add another constructor to keep binary compatibility.

Andrapyre:
I've adjusted both PRs to keep bincompat. Build should be passing now.

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -944,11 +944,46 @@ private[magnolia1] object CompileTimeState {
...
+    */
+  def withValueEvaluator[A](a: => A): CallByNeed[A] = new CallByNeed(() => a, () => true)
+}
+
+// Both params are later nullified to reduce overhead and increase performance.
+// The supportDynamicValueEvaluation is passed as a function so that it can be nullified. Otherwise, there is no need for the function value.
+final class CallByNeed[+A] private (private[this] var eval: () => A, private var supportDynamicValueEvaluation: () => Boolean)
+    extends Serializable {
+
+  def this(eval: () => A) = this(eval, () => false)
```
adamw, line 967: that's required for bin compat, right? can you add a comment, so that when we'll work on v2 at some point the future, we'll know what to simplify :)

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -944,11 +944,46 @@ private[magnolia1] object CompileTimeState {
   }
 }
 
-object CallByNeed { def apply[A](a: => A): CallByNeed[A] = new CallByNeed(() => a) }
-final class CallByNeed[+A](private[this] var eval: () => A) extends Serializable {
+object CallByNeed {
+
+  /** Initializes a class that allows for suspending evaluation of a value until it is needed. Evaluation of a value via `.value` can only
```
adamw, line 949: that's different from previous semantics, where the value was memoized in a lazy val (and I think it still is) - so the value can be accessed via `.value` multiple times, but evaluation happens once

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -944,11 +944,46 @@ private[magnolia1] object CompileTimeState {
...
+}
+
+// Both params are later nullified to reduce overhead and increase performance.
+// The supportDynamicValueEvaluation is passed as a function so that it can be nullified. Otherwise, there is no need for the function value.
+final class CallByNeed[+A] private (private[this] var eval: () => A, private var supportDynamicValueEvaluation: () => Boolean)
+    extends Serializable {
+
+  def this(eval: () => A) = this(eval, () => false)
+
+  val valueEvaluator: Option[() => A] = {
```
adamw, line 969: shouldn't this be a `lazy val` as well? otherwise we compute this upfront when the class is constructed, not on-demand. Or maybe we can make this a `def`, which would simply call `eval()` if it's not-null, and if it's already nulled, would throw an exception saying that you can only access the value using either .value, or .valueEvaluator, but not both?

core/src/main/scala/magnolia1/magnolia.scala
Andrapyre: Done.

core/src/main/scala/magnolia1/magnolia.scala
Andrapyre: To take a step back, this class involves the following:
1. Initial data is nullified as much as possible to keep the class lightweight.
2. Cannot change fields or class structure because of bincompat
3. The value evaluator is unused in other use cases not involving default params and is therefore adding overhead that needs to be reduced as much as possible

Ideally (something for v2), there would be two classes - something like `CallByNeedStatic` and `CallByNeedDynamic` that each extend a `CallByNeed` trait, each containing a `lazy val value` and the second containing an additional `def evaluateValue()`, as you suggested. 

Since the ideal option is not possible due to backwards compatibility requirements, I've tried to do the next best thing:
1. Add an additional param, `supportDynamicValueEvaluation`, which will ensure that the value evaluator is `Some` when the flag is set to true. The param is passed as a function so that it can later be nullified. If it is false, then the evaluator is set to `None` so that overhead is reduced.
2. I considered and ruled out making both values lazy or making `valueEvaluator` into a `def`, since both of these would involve throwing intransparent initialization errors if one `lazy val` is initialized before the other or the `lazy val` is called before the `def`. I considered it too bug prone and not user-friendly. Apis should be transparent and easy for end-users to reason about and either behavior would violate that expectation.

core/src/main/scala/magnolia1/magnolia.scala
Andrapyre: Semantics here don't change, but perhaps the comment can be worded in a way that's clearer? Previous functionality also has evaluation/computation only occurring once, but of course it can be called multiple times.

adamw: Yeah I think it's the wording that caught me - usage of "can" vs. "will"

core/src/main/scala/magnolia1/magnolia.scala
adamw: Yeah I agree that two classes would be best, and that's what we'll eventually do. Re-reading the code initializing `valueEvaluator` on construction should work fine. Thanks for the explanation.

adamw:
Thanks for the PR!

PR #559, opened 2025-01-11
uncomment and fix java annotation tests

hughsimpson:
I'm not sure if this is somehow a really stupid way of doing it, but it seems to fix the tests, as well as some of my downstream issues *. Changing the type of the collected annotations without changing the api is probably dumb, but I figured it signalled intent without breaking any compatibility. It's not really necessary is it. Well anyway let me know your thoughts.

\* I say 'some', because there's still an issue with scala annotations - I think it happens if you annotate both a case class and a parent trait with the same annotation, but with this change I don't have to worry about that problem because I can just use the upstream java annotations instead of trying to work around the issue. Might make a follow-up at some point...

### DIFF
--- a/core/src/main/scala/magnolia1/macro.scala
+++ b/core/src/main/scala/magnolia1/macro.scala
@@ -178,14 +178,14 @@ object Macro:
 
     private val tpe: TypeRepr = TypeRepr.of[T]
 
-    def anns: Expr[List[Any]] =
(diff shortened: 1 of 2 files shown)


hughsimpson:
Apologies, had missed that formatting hadn't been applied. fixed and pushed

adamw:
Well if the tests pass, it works! :)

PR #560, opened 2025-01-13
bump deps and fix java annotations

hughsimpson:
'forward-porting' https://github.com/softwaremill/magnolia/pull/559

Since v2 is presumably intended to encapsulate breaking changes, I figured 'hey, let's make all these `Any` decls `Annotation` decls since that's nicer'.

I also bumped some deps so that I could build locally on Java 21

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import com.softwaremill.UpdateVersionInDocs
 import com.softwaremill.SbtSoftwareMillCommon.commonSmlBuildSettings
 import com.softwaremill.Publish.{updateDocs, ossPublishSettings}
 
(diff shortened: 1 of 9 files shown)


hughsimpson:
I don't really know anything about the mima stuff but imagine that the failure there is inevitable and ignorable? I made this because I saw the ticket about V2 and thought it a good opportunity to make the extra API changes...

adamw:
@hughsimpson thanks! yeah there's a couple of other changes as well waiting for porting, so that at some point we can get v2 up to speed and ship ;)

adamw:
since it's v2 mima fails (as there are other bin-incompat changes), so that's expected

PR #567, opened 2025-02-04
Add def `rawConstruct(fieldValues: Array[Any]): Type` to reduce allocations and CPU usage

### LINKED ISSUE #420: Add `def rawConstruct(fieldValues: Array[Any]): Type` for perf
When we already have an `Array[Any]` in our derivation, it would be more performant to use `def rawConstruct(fieldValues: Array[Any]): Type` than `def rawConstruct(fieldValues: Seq[Any]): Type`.

plokhotnyuk:
Fix https://github.com/softwaremill/magnolia/issues/420

### DIFF
--- a/core/src/main/scala/magnolia1/impl.scala
+++ b/core/src/main/scala/magnolia1/impl.scala
@@ -62,10 +62,13 @@ object CaseClassDerivation:
         typeAnnotations
       ):
     def construct[PType: ClassTag](makeParam: Param => PType): A =
-      product.fromProduct(Tuple.fromArray(parameters.map(makeParam).to(Array)))
+      product.fromProduct(Tuple.fromIArray(parameters.map(makeParam)))
 
-    def rawConstruct(fieldValues: Seq[Any]): A =
-      product.fromProduct(Tuple.fromArray(fieldValues.to(Array)))
+    override def rawConstruct(fieldValues: Seq[Any]): A =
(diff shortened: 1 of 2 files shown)


adamw:
thanks - mima's unhappy, though

plokhotnyuk:
It seems MiMa check fails on checking _forward_ binary compatibility as for the patch release.

How to increase minor version to 1.4.0?

adamw:
Hm I thought MiMA only checks backwards binary compatibility. What does it hurt that a new method is added ... if the new version is on the classpath, everything should work just fine?

plokhotnyuk:
It will be safe if implementations of `rawConstruct` should be only in the internals of the magnolia library.

I can add an exclude rule if my assumption about only internal implementations is true.

adamw:
Ah right, yeah I get the problem now. That is indeed backwards-incompatible - for any external implementations of `CaseClass`. Shouldn't we then keep the `Seq` variant abstract, in the `Array` variant delegate to `Seq` by default, and in internal implementations "do the right thing" - override both and delegate the way we want to?

plokhotnyuk:
Done, please approve the workflow run

adamw:
Cool, I'll cut a new release, or do you have any more changes?

plokhotnyuk:
Thanks a lot! Please cut one, and I will continue measurements

PR #568, opened 2025-02-04
More efficient default value evaluator

plokhotnyuk:
Currently for each evaluator function call the hash map lookup with subsequent `flattern` and `flatmap` calls is happening:

image

This PR fixes that.


plokhotnyuk:
It gives great speed up for some protocols like OpenRTB where almost all fields have defaults.

Below are flame graphs for CPU cycles with evaluation of defaults highlighted by magenta: 

#### Before
image

#### After
image

adamw:
Thanks :)

plokhotnyuk:
Thanks for your support in improving performance!

Can I ask for cutting a new release? 

As you saw on flame graphs my changes are really helping for case classes which are full of defaults, and users would love to use them in a stable release as soon as possible.

adamw:
Already in progress :)

Issue #565, opened 2025-01-27
Failure to derive typeclass for large sum types on Scala 3


ghostdogpr:
If you have a sealed trait with lots of subtypes (~1000 in my case, but the issue happens with much less), Magnolia runs into `max-inlines` issues or even triggers a compiler stack overflow.

This comes from the fact that the `subtypesFromMirror` [method](https://github.com/softwaremill/magnolia/blob/scala3/core/src/main/scala/magnolia1/impl.scala#L174) is not tail-recursive. It folds over the list of subtypes and increases the number of nested inline function calls and the stack depth proportionally to the number of subtypes.

See [this example in Caliban](https://github.com/ghostdogpr/caliban/blob/eb6575a31d9cb445077ec8138b4d277246dbcb5c/core/src/main/scala-3/caliban/schema/SchemaDerivation.scala#L42) on how it can be made tail-recursive. 

On top of that, `subtypesFromMirror` calls `.distinctBy(_.typeInfo).sortBy(_.typeInfo.full)` [on every iteration](https://github.com/softwaremill/magnolia/blob/f4945a2a41f2a68bcdf1a4b6dcaff24a2e225bd6/core/src/main/scala/magnolia1/impl.scala#L213), which is very inefficient with large list of subtypes. This should be moved at the end of the loop.

adamw:
Indeed :) Would you be able maybe to work on a fix?

ghostdogpr:
> Indeed :) Would you be able maybe to work on a fix?

Not right now, I'm in the middle of a large Scala 3 migration and I ended up not using Magnolia since I could do it [more simply](https://gist.github.com/ghostdogpr/6f2ca0939c67765a0657a255ed653765). But I thought I'd share my findings anyway, hopefully it's relatively simple!

adamw:
Sure thing, thanks :)

adamw:
Let's reopen if #573 doesn't solve the problem

PR #573, opened 2025-02-15
Fix failure to derive typeclass for large sum types on Scala 3

ex-ratt:
This should resolve issue #565 by making `subtypesFromMirror` tail-recursive. It doesn't use the `@tailrec` annotation however, just as the example code linked in the issue, because this isn't possible with inline methods. I hope that it does indeed fix the problem... I basically just followed the suggestions from the issue.


adamw:
Let's try :)

Issue #574, opened 2025-02-16
Avoid recursive inlining for product types


MartinHH:
Somewhat related to / inspired by #565 :

## Motivational example

```scala
//> using dep "com.softwaremill.magnolia1_3::magnolia:1.3.13+0-44422700+20250215-1851-SNAPSHOT"

import magnolia1._

trait TypeNameInfo[T] {
  def name: TypeInfo
}

object TypeNameInfo extends ProductDerivation[TypeNameInfo]:
  def join[T](ctx: CaseClass[TypeNameInfo, T]): TypeNameInfo[T] =
    new TypeNameInfo[T]:
      def name: TypeInfo = ctx.typeInfo

  given fallback[T]: TypeNameInfo[T] =
    new TypeNameInfo[T]:
      def name: TypeInfo = TypeInfo("", "Unknown Type", Seq.empty)

case class BigCaseClass(
  a1: Int, b1: Int, c1: Int, d1: Int, e1: Int, f1: Int, g1: Int, h1: Int, i1: Int, j1: Int,
  k1: Int, l1: Int, m1: Int, n1: Int, o1: Int, p1: Int, q1: Int, r1: Int, s1: Int, t1: Int,
  u1: Int, v1: Int, w1: Int, x1: Int, y1: Int, z1: Int,
  a2: Int, b2: Int, c2: Int, d2: Int, e2: Int, f2: Int, g2: Int, h2: Int, i2: Int, j2: Int,
  k2: Int, l2: Int, m2: Int, n2: Int, o2: Int, p2: Int, q2: Int, r2: Int, s2: Int, t2: Int,
  u2: Int, v2: Int, w2: Int, x2: Int, y2: Int, z2: Int
) derives TypeNameInfo
```
compilation fails with
```scala
[error] Maximal number of successive inlines (32) exceeded,
[error] Maybe this is caused by a recursive inline method?
[error] You can use -Xmax-inlines to change the limit.
[error] ) derives TypeNameInfo
```

## Proposed improvement: use `summonAll` instead of recursion

When compiling with Scala 3.4.0 or higher (thanks to [this commit](https://github.com/scala/scala3/commit/d71ca06419a69b38b7ae0e932e47bc13577a8200)), `scala.compiletime.summonAll` does not rely on recursive inlining anymore.

[...]

PR #575, opened 2025-02-16
Avoid recursive inlining (for products)

MartinHH:
Leverage summonAll to avoid recursive inlining as proposed in #574

So far, this is just a proof of concept.

Various tests are still failing - most of them are related to error messages of compilation errors not being as expected, but
one (["derive instances for types with refined types if implicit provided"](https://github.com/softwaremill/magnolia/blob/44422700e0fbfc7ccb8dc6ff54880edf181e7cdf/test/src/test/scala/magnolia1/tests/OtherTests.scala#L41)) seems a bit more serious (compilation fails when it shouldn't).

### DIFF
--- a/core/src/main/scala/magnolia1/impl.scala
+++ b/core/src/main/scala/magnolia1/impl.scala
@@ -96,6 +96,18 @@ object CaseClassDerivation:
       } { params => product.fromProduct(Tuple.fromArray(params)) }
     }
 
+  private trait ParamFactory[TC[_], P]:
+    def tc: SerializableFunction0[TC[P]]
+    def isType(v: Any): Boolean
+
+  private object ParamFactory:
+    inline given paramFactory[TC[_], P]: ParamFactory[TC, P] =
+      new ParamFactory[TC, P]:
+        def tc: SerializableFunction0[TC[P]] =
+          new SerializableFunction0[TC[P]]:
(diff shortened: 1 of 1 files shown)


MartinHH:
> ["derive instances for types with refined types if implicit provided"](https://github.com/softwaremill/magnolia/blob/44422700e0fbfc7ccb8dc6ff54880edf181e7cdf/test/src/test/scala/magnolia1/tests/OtherTests.scala#L41)

Turns out that one doesn't fail when changing the project's scala version to 3.4.0 (and the unexpected error messages from the other failing tests are also much closer to the expected ones).

Maybe it would make sense to postpone this change until the next LTS version (3.7.0) is out and Magnolia is migrated to that (?).

core/src/main/scala/magnolia1/impl.scala
```diff
@@ -96,6 +96,18 @@ object CaseClassDerivation:
       } { params => product.fromProduct(Tuple.fromArray(params)) }
     }
 
+  private trait ParamFactory[TC[_], P]:
```
MartinHH, line 99: This is where one would need to add an `@implicitNotFound` annotation in order to fix the failing tests related to (compilation) error messages.

Something along the lines of:
```
@implicitNotFound("No given instance of type ${TC} (for type ${P}) was found")
```

Issue #294, opened 2021-05-18
Scala 3 caveats


ghostdogpr:
A few other things I discovered and that you might want to mention it somewhere: Scala 3 typeclass derivation doesn't support a few things that Magnolia supported with Scala 2:
- value classes
- nested sealed traits

Opaque types also don't have mirrors.
And I think default values can't be obtained from the macro yet.

adamw:
I suppose a good idea here would be to move all of the magnolia-scala2 tests into magnolia-scala3, and see which ones fail. Mark them as such, but keep them around as a good documentation for further work.

PR #589, opened 2025-03-20
value class support and typeclass paramaterisation

hughsimpson:
Two things have been bugging me about magnolia in scala 3:

- the inability to paramaterise typeclass construction with an auxiliary typeclass
- lack of support for value classes

This goes some of the way towards addressing those issues. Since it changes derivation signatures I decided to implement this as 'more capable' variants of the derivation traits, rather than breaking all the bin compat.

Value class support is implemented by providing an appropriate mapping on the derivation but is a bit clunky as it stands, and the macros don't support: non-product (i.e. non-case class) wrappers, private constructors for 'extraction' typeclass, or private accessors for 'wrapping' typeclasses; although I think the last two are achievable with judicious modification of the macro, and the first _might_ be, although I'm not sure how, and it would probably break the signatures here...

See https://github.com/geirolz/cats-xml/pull/251 for motivation 

Obviously I'm more than happy to amend aspects of this according to feedback, so arguably I should've opened this as a draft WIP, but there aren't any further changes I'd particularly planned to make initially (other than naming, I see now. Left in 'Serde' in some names, but that aside...). Feedback welcome

### DIFF
--- a/examples/src/main/scala/magnolia1/examples/show_paramaterised.scala
+++ b/examples/src/main/scala/magnolia1/examples/show_paramaterised.scala
@@ -1,0 +1,110 @@
+package magnolia1.examples
+
+import magnolia1._
+import magnolia1.parameterised._
+
+/** shows one type as another, often as a string
+  *
+  * Note that this is a more general form of `Show` than is usual, as it permits the return type to be something other than a string.
+  */
+trait ParamaterisedShow[Out, T] extends Serializable {
+  def show(value: T): Out
+  def contramap[S](fn: S => T): ParamaterisedShow[Out, S] = (value: S) => show(fn(value))
+}
+
+trait ShowConfig[T] {
+  def hiddenFields: Set[String]
+}
+object ShowConfig {
+  given default[T]: ShowConfig[T] = new ShowConfig[T] {
+    override def hiddenFields: Set[String] = Set.empty
+  }
+}
+
+trait ParamaterisedGenericShow[Out] extends ParamaterisedAutoDerivation[[X] =>> ParamaterisedShow[Out, X], ShowConfig] {
+
+  def joinElems(typeName: String, strings: Seq[String]): Out
+  def prefix(s: String, out: Out): Out
+
+  /** creates a new [[Show]] instance by labelling and joining (with `mkString`) the result of showing each parameter, and prefixing it with
+    * the class name
+    */
+  def join[T: ShowConfig](ctx: CaseClass[Typeclass, T]): ParamaterisedShow[Out, T] = { value =>
+    if ctx.isValueClass then
+      val param = ctx.params.head
+      param.typeclass.show(param.deref(value))
+    else
+      val paramStrings = ctx.params.map { param =>
+        val attribStr =
+          if (param.annotations.isEmpty && param.inheritedAnnotations.isEmpty)
+            ""
+          else {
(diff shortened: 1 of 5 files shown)


examples/src/main/scala/magnolia1/examples/show_paramaterised.scala
```diff
@@ -0,0 +1,110 @@
...
+    params.mkString(s"$typeName(", ",", ")")
+
+  given ParamaterisedShow[String, String] = identity(_)
+  given ParamaterisedShow[String, Int] = _.toString
+  given ParamaterisedShow[String, Long] = _.toString + "L"
+  given ParamaterisedShow[String, Boolean] = _.toString
+  given [A](using A: ParamaterisedShow[String, A]): ParamaterisedShow[String, Seq[A]] =
+    _.iterator.map(A.show).mkString("[", ",", "]")
+
+  override def handleAnyVal: [T <: AnyVal, S] => (
```
hughsimpson, line 105: This is fiddly to implement, but still better than writing the macros yourself... I imagine it can be a whole lot nicer than this

hughsimpson:
The structure here is probably a little wonky... Thinking about it, the value class support should probably be some kind of optional mix-in...

hughsimpson:
Am I the only one who misses these capabilities, or is there an approach for them that I've missed, or...? I'll push up some more strenuous tests as soon as I have the time, but I'd like to get confirmation that I'm not just completely mad and that this is a reasonable direction to take

Edit: maybe I shouldn't be trying to combine this into one pr? I dunno. Feedback very welcome.

adamw:
Quite the contrary, there are often questions about value classes. However, the state of this feature isn't that rosy: the related issues are https://github.com/softwaremill/magnolia/issues/296 and https://github.com/softwaremill/magnolia/issues/294. At some point we started an effort to release magnolia-scala3 v2, with value classes support: https://github.com/softwaremill/magnolia/pull/443, but unfortunately it stalled, and even more unfortunately, the v1 & v2 branches diverged. So I guess it's still our plan to re-create v2 or port the changes to the existing v2, but I don't really know when ;)

As for your PR, as I understand it introduces a parallel derivation mechanism? I guess it would be better than just to work on a v2 version. Also, I think using `Mirror`s can have poor performance, so going macro-only might yield better results.

Finally, you write:

> the inability to paramaterise typeclass construction with an auxiliary typeclass

can you share an example of what the problem here is?

hughsimpson:
Thanks for the feedback. Tbh I opened this pr too early and there were unresolved issues even to my mind once I read it through - will try to get back to it soon, but will convert to a draft in the meantime. 

The specific use-case I had in mind re paramaterisation was for cats-xml, where the pattern is to instantiate an implicit declaring certain properties - such as field foo should be an annotation, field bar should be a node, field baz should be text content - and use that in the derivation to generate the appropriate serde. Although I've run into a similar issue once where the usage wsa intended to specify the discriminator for some JSON serdes being derived for spray-json, and I pretty much couldn't do it without the derive implementation knowing about the traits in advance and this would've helped there. I'll try to write some tests to show this sort of use case

I'll take a look at V2 that's probably a better direction.

adamw:
Ah yes, having (failing) tests would be great.

As for V2, I'm also hoping to include some ideas by @MateuszKubuszok (see https://www.youtube.com/watch?v=h9NdXLTZkGk) to make Magnolia not only feature-full, but also fast :)

PR #595, opened 2025-05-13
Unroll recursive inline loops

### LINKED ISSUE #574: Avoid recursive inlining for product types

road21:
I have tried to avoid unrolling boilerplate (like in https://github.com/scodec/scodec/pull/586, https://github.com/circe/circe/pull/2341), but that deferred inline approach worked only for `CaseClassDerivation`.
Probably, because `SealedTraitDerivation` contains `inline def deriveSubtype` that plays bad with other deferred inlines.

Ugly, but closes https://github.com/softwaremill/magnolia/issues/574


core/src/main/scala/magnolia1/impl.scala
```diff
@@ -107,29 +140,65 @@ object CaseClassDerivation:
     inline erasedValue[(Labels, Params)] match
       case _: (EmptyTuple, EmptyTuple) =>
         Nil
+      case _: (
```
adamw, line 146: I don't think I understand, why are we matching only empty and 16-tuples? What about all the others?

road21: We match 
- empty
- 16
- 4 
- 1

These arities allows to inline tuples up to 27 * 16 = 432 fields:
(in worst case for 431 field: (26 inlines) * (by 16) + (3 inlines) * (by 4) + (3 inlines) * (by 1))

KacperFKorban: Aren't there a few more `inline` calls before the first call to `paramsFromMaps`? I think that the numbers will be slightly different then.
Also I'm wondering if it won't be a bit weird that the set of supported sizes has "holes" i.e. some unsupported sizes are smaller than some supported sizes. That is an extreme example though, so I would say that we should at least handle 0-255 (max number of arguments to a method on the JVM).

core/src/main/scala/magnolia1/impl.scala
```diff
@@ -96,6 +96,39 @@ object CaseClassDerivation:
...
+      CallByNeed.createLazy(tc),
+      CallByNeed.createValueEvaluator(d),
+      repeated,
+      annotations,
+      inheritedAnnotations,
+      typeAnnotations,
+      idx
+    )
+
   inline def paramsFromMaps[Typeclass[_], A, Labels <: Tuple, Params <: Tuple](
```
KacperFKorban, line 135: It would be useful to add a comment explaining why the function uses this unrolled recursion.

KacperFKorban (commented):
Overall, the changes look good to me. I would just make sure that the set of supported sizes is well defined.

core/src/main/scala/magnolia1/impl.scala
road21: > Aren't there a few more inline calls before the first call to paramsFromMaps? I think that the numbers will be slightly different then.

Yes, that is true

I have added tests for 254 field (compiler error about platform restriction for 255 fields)
There is no holes in 1-254 range, 254 is a worst case for 16,4,1 arities

The only way to support sizes without holes - arities without holes )
I can do that, but it looks like overkill

adamw:
Thanks @KacperFKorban for the review and @road21 for the changes :) 

there are some test failures, though - stack overflow in the compiler

road21:
> there are some test failures, though - stack overflow in the compiler

I can reproduce it only when running `clean; exit` before compilation
Increasing stack size doesn't help, so it looks like a compiler bug

I will try to minimize it and find workaround

road21:
I have minimized it to https://github.com/road21/sbt-compiler-stackoverflow-phantom-bug

The problem is that stackoverflow is phantom and I don't know how to make reproducible example (this example fails on my machine randomly and compiles on my colleagues machines)

Let's try to increase Xss on pipelines...

adamw:
I guess it worked - I ran CI 3 times ;) Thanks!

Issue #627, opened 2026-02-17
Support covariant typeclass families


nigredo-tori:
This issue is motivated by the Caliban library, which defines a `Schema` typeclass like this:
```scala
abstract class Schema[-R, T]
```
`R` can be different for different `T`, but we can combine heterogenous instances. For example, for some case class with fields that have corresponding instances `Schema[R1, Field1]` and `Schema[R2, Field2]` we can build a `Schema[R1 & R2, SomeCaseClass]`. However, as it stands right now, we can't easily derive it using Magnolia, because we have `Schema.Typeclass` defined as a `Schema[Any, *]`, so that is what `gen` produces.

The way Caliban "solves" this is by providing a `GenericSchema` [helper](https://ghostdogpr.github.io/caliban/docs/schema.html#effects) to the library user, with its own specialized `Typeclass`. It gets awkward to use without giving up the way ZIO code tends to be structured. I won't get into details, suffice to say that I'd like to only have to rely on `Schema.gen` for derivation.

I imagine that everywhere where we use `Typeclass` (e.g. [here](https://github.com/softwaremill/magnolia/blob/scala2/core/src/main/scala/magnolia1/magnolia.scala#L414) that is what `typeConstructor` is bound to), we can instead use a narrower type. It's a bit awkward to explain, so I'll add a PR for Scala 2 shortly.

Does this change fit this library, or is it specifically geared towards invariant typeclasses? I feel like with the rise of ZIO we are likely to see more of covariance everywhere, so cases like this could get somewhat more common.

nigredo-tori:
Hm. It's kind of difficult to do this for Scala 3, since `join` and `split` signatures are [set in stone](https://github.com/softwaremill/magnolia/blob/993165257c81b869ab4e559f05f58e3e903e3bef/core/src/main/scala/magnolia1/magnolia.scala#L17), so I don't see how the calling code could define something like [this](https://github.com/softwaremill/magnolia/blob/scala2/examples/src/main/scala/magnolia1/examples/collectFields.scala#L24). It may be possible to make them `inline`...

[...]

PR #628, opened 2026-02-17
Support covariant typeclass families in Scala 2

nigredo-tori:
See #627


core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -6,11 +6,14 @@ import scala.annotation.{compileTimeOnly, tailrec}
 import scala.collection.mutable
 import scala.language.higherKinds
 import scala.reflect.macros._
+import scala.reflect.ClassTag
 
 /** the object which defines the Magnolia macro */
 object Magnolia {
   import CompileTimeState._
 
+  private[Magnolia] type ConstNothing[a] = Nothing
```
nigredo-tori, line 17: A "typeclass" that is a subtype of any typeclass.

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -381,37 +385,36 @@ object Magnolia {
...
-        val functionName = c.freshName(TermName("partialAssignments"))
-        val arrayVal = c.freshName(TermName("arr"))
+      def constructPartialSubtypesValFunction(typeclasses: List[(Type, Tree)], arrayElementType: Type): (TermName, Tree) = {
+        val functionName = c.freshName(TermName("partialSubtypes"))
         val startVal = c.freshName(TermName("start"))
-        val assignments = typeclasses.zipWithIndex.map { case ((subType, typeclass), idx) =>
+        val subtypeObjects = typeclasses.zipWithIndex.map { case ((subType, typeclass), idx) =>
           val symbol = subType.typeSymbol
           val (annotationTrees, inheritedAnnotationTrees) = annotationsOf(symbol)
-          q"""$arrayVal($startVal + $idx) = $SubtypeObj[$typeConstructor, $genericType, $subType](
```
nigredo-tori, line 414: Remove "hard" type specifications where possible.

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -565,14 +565,19 @@ object Magnolia {
             ((((param, idx), default), (annList, inheritedAnnList)), typeAnnList) <-
               paramsWithIndex zip defaults zip annotations zip typeAnnotations
           )
-            yield param.compile(paramsVal, idx, Some(default), annList, inheritedAnnList, typeAnnList)
+            yield param.compile(idx, Some(default), annList, inheritedAnnList, typeAnnList)
+        }
+
+        val paramsValDef = {
+          val method = TermName(if (isReadOnly) "readOnlyParams" else "params")
+          q"$PartsObj.$method[$typeConstructor, $genericType](..$paramsItems)"
```
nigredo-tori, line 577: When building `paramsVal`, we simultaneously let the typer calculate the narrowest subtype of the original `Typeclass` we can fit over the params. We can then use it instead of the original `Typeclass`.

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -988,3 +994,20 @@ final class CallByNeed[+A] private (private[this] var eval: () => A, private var
...
 }
+
+final class Parts[F[+_[_]], Tc[_]]private[Parts] (val array: Array[F[Tc]]) extends Serializable {
+  type Typeclass[a] = Tc[a]
+}
+
+/** Helpers to guide `Param`/`Subtype` array types and to provide access to the resulting typeclass.
+  */
+object Parts {
+  def params[TcLower[_], T]: PartiallyApplied[Param[*[_], T],  TcLower] = new PartiallyApplied[Param[*[_], T],  TcLower]()
```
nigredo-tori: This is the magic that lets the typer choose a narrower typeclass, while still being constrained by the original `Typeclass` (`TcLower` here).

adamw:
It seems the build fails?

nigredo-tori:
> It seems the build fails?

Not because of anything in the PR:

> The job has exceeded the maximum execution time while awaiting a runner for 24h0m0s

It seems that the check uses the runners assigned to your repository, and I can't restart it (which makes sense). I see that you've already tried to restart it with the same result. My best guess is that this is caused by ubuntu-20.04 images [being discontinued](https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/#ubuntu-20-image-is-closing-down) for GitHub Actions. You've [changed the images](https://github.com/softwaremill/magnolia/commit/ea4dd59213e2e21c0d5a71ebb731a3a116d33504) for the `scala3` branch but not for `scala2`.

adamw:
Ah of course, thanks for pointing that out. I think the build should be updated now

core/src/main/scala/magnolia1/magnolia.scala
adamw: these are great comments - can you maybe put them as comments on code? they help to understand what's going on :)

nigredo-tori: Done. I've also done some cleanup.

adamw:
Thanks! :)

PR #631, opened 2026-02-24
Update Ubuntu images in workflows

nigredo-tori:
https://github.com/softwaremill/magnolia/pull/628#issuecomment-3952588822

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
   ci:
     # run on external PRs, but not on internal PRs since those will be run by push to branch
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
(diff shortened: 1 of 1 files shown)


nigredo-tori:
Hm. I gave this a try, now the runners start, at least. It seems like `matrix.java` does not get resolved because in https://github.com/softwaremill/magnolia/commit/eb2764cc2cbbc585ed3f7d5c49bb3d30c45fca78 we haven't specified it, and we haven't switched all the way to softwaremill/github-actions-workflows thing like [in scala3](https://github.com/softwaremill/magnolia/blob/scala3/.github/workflows/ci.yml#L13). I'm familiar with neither this setup, nor with GitHub workflows in general, and I'm a bit too burned out to figure it out right now. So I'm closing this PR, but anybody else is welcome to continue fixing this.

Issue #634, opened 2026-03-12
unused warnings after macro expansion for sealed traits


road21:
When running tests in this repo, you get warnings:

```scala
Show.gen[TestEntry].show(TestEntry("a", "b"))

// parameter value start$macro$117 in method partialSubtypes$macro$116 is never used
// [warn]       val res = Eq.gen[Entity].equal(Person("John Smith", 34), Person("", 0))
```

Looks like the problem is that there is an unused variable `start` in generated code ([source](https://github.com/softwaremill/magnolia/blob/8cf9508ea2b0da9baecc78fd5fde81abb9b1418c/core/src/main/scala/magnolia1/magnolia.scala#L430)):

```scala
    q"""def $functionName($startVal: $IntTpe) = {
        $PartsObj.subtypes[$typeConstructor, $genericType](..$subtypeObjects)
    }"""
```

The problem is that this warning will be thrown in any user code.
I guess to prevent problems like this we should turn on fatal warnings at least for tests

nigredo-tori:
Can this be closed?

adamw:
I believe so!

PR #635, opened 2026-03-12
remove unused val in generated code

road21:
Closes https://github.com/softwaremill/magnolia/issues/634


road21:
I guess this relates to changes from https://github.com/softwaremill/magnolia/pull/628

@nigredo-tori could you also take a look, please?

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -426,12 +424,7 @@ object Magnolia {
             (t: $genericType) => t.asInstanceOf[$subType]
           )"""
         }
-        val tree =
-          q"""def $functionName($startVal: $IntTpe) = {
-                $PartsObj.subtypes[$typeConstructor, $genericType](..$subtypeObjects)
-              }"""
-
-        (functionName, tree)
+        q"""$PartsObj.subtypes[$typeConstructor, $genericType](..$subtypeObjects)"""
```
nigredo-tori: The reason we have a bunch of functions here is that for some niche use cases (sealed traits with hundreds of subtypes) we used to hit Java method size limit. See softwaremill/magnolia#485 where this was introduced. This PR breaks that use case, even though we don't have a test for it.

You could have just removed the parameter, and massaged the `combinations` accordingly. But, since you've already done some refactoring, I suggest changing this line to
```suggestion
        q"""(() => $PartsObj.subtypes[$typeConstructor, $genericType](..$subtypeObjects))()"""
```
Which looks weird, but saves us the hassle of dealing with named functions.

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -409,9 +409,7 @@ object Magnolia {
           $impl
       """
 
-      def constructPartialSubtypesValFunction(typeclasses: List[(Type, Tree)]): (TermName, Tree) = {
-        val functionName = c.freshName(TermName("partialSubtypes"))
-        val startVal = c.freshName(TermName("start"))
+      def constructPartialSubtypesValFunction(typeclasses: List[(Type, Tree)]): Tree = {
```
nigredo-tori: This no longer produces functions, so we can rename it to `constructPartialSubtypesVal`

nigredo-tori (changes requested):
LGTM other than the function thing.

core/src/main/scala/magnolia1/magnolia.scala
road21: renamed

core/src/main/scala/magnolia1/magnolia.scala
road21: I have added test - you are right, my fix breaks this case, but your fix actually helps (to be honest I don't fully understand how)

test/src/test/scala/magnolia1/tests/tests.scala
```diff
@@ -889,6 +889,11 @@ class Tests extends munit.FunSuite {
       assertEquals(res, "Bar(foo{MyAnnotation(2),MyAnnotation(1)}=foo,bar{MyAnnotation(2),MyAnnotation(1)}=bar)")
     }
 
+    test("construct a Show instance for huge enums") {
+      val res = Show.gen[HugeEnum].show(HugeEnum.Case42)
```
nigredo-tori, line 893: Thank you for adding a test for this!

core/src/main/scala/magnolia1/magnolia.scala
nigredo-tori: The root issue is that if we have **a lot** of code in one **JVM/bytecode** method, we hit the limit on the size of this method. If we naively build the whole instance for a massive type without the existing workaround, then all of that building happens in one place, and gets compiled as a part of one bytecode method.

To avoid this, #485 took parts of the logic, and moved them to separate functions. When compiling to bytecode, each of those functions is compiled to its own bytecode method. As a result, the remainder of the logic is compiled to significantly less bytecode, which has better chance of not hitting the limit.

The change I've proposed here simply makes those functions anonymous:
```scala
() => $PartsObj.subtypes[$typeConstructor, $genericType](..$subtypeObjects)
```
, and immediately calls them:
```scala
(() => $PartsObj.subtypes[$typeConstructor, $genericType](..$subtypeObjects))()
```
The result is for all intents and purposes the same, other than the fact that we no longer need to emit declarations for the functions in a separate location, or deal with their names.

nigredo-tori (approved):

road21:
@adamw Could you please take a look?

adamw:
@road21 Sure, merged, thanks! :)

PR #637, opened 2026-03-22
Add genNarrow

nigredo-tori:
See #627.

It's difficult to narrow recursive instances, especially in Scala 2 where we don't constrain `join` and `split` signatures. The most robust solution (although not the most convenient) is to give the users a way to manually specify the desired typeclass.

This PR adds a new macro, `genNarrow`, that does just that. It reuses most of the implementation of `gen`, but replaces uses of `Typeclass` with a specified `* -> *` HKT.

### DIFF
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -46,8 +46,28 @@ object Magnolia {
     * [[SealedTrait]], like so, <pre> &lt;derivation&gt;.split(&lt;sealedTrait&gt;): Typeclass[T] </pre> so a definition such as, <pre> def
     * split[T](sealedTrait: SealedTrait[Typeclass, T]): Typeclass[T] = ... </pre> will suffice, however the qualifications regarding
     * additional type parameters and implicit parameters apply equally to `split` as to `join`.
+    *
+    * `gen` will make an effort to keep the resulting value's type as narrow as possible, which can be useful for typeclass families. As a
+    * contrived example, when configured with `type Typeclass[x] = Either[Any, x]` and corresponding `join` and `split`, `gen` may derive an
+    * `Either[String, T]` if the target type and available instances allow it. To take advantage of it, one has to define `join` and `split`
+    * in such a way that they propagate narrowed typeclasses as appropriate:
+    *
+    * {{{
+    * def join[L, T](caseClass: CaseClass[Either[L, *], T]): Either[L, T]
+    * }}}
+    *
+    * Note that narrowing is not perfect, and it will fail for (mutually) recursive target types, producing `Typeclass[T]` instead of a more
+    * specific variant. To handle such cases, please refer to the [[genNarrow]] macro which supports choosing a specific typeclass family member.
     */
-  def gen[T: c.WeakTypeTag](c: whitebox.Context): c.Tree = Stack.withContext(c) { (stack, depth) =>
+  def gen[T](c: whitebox.Context)(implicit T: c.WeakTypeTag[T]): c.Tree =
+    genImpl(c)(TypeConstructor.fromTypeclass(c), T)
+
+  /** Like [[gen]], but instead of `Typeclass[T]` this produces `Tc[T]` for the specified `Tc[_]`.
+    */
+  def genNarrow[Tc[_], T](c: whitebox.Context)(implicit Tc: c.WeakTypeTag[Tc[_]], T: c.WeakTypeTag[T]): c.Tree =
+    genImpl(c)(TypeConstructor.fromTag[Tc](c)(Tc), T)
+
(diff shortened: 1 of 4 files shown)


core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -46,8 +46,28 @@ object Magnolia {
...
     * additional type parameters and implicit parameters apply equally to `split` as to `join`.
+    *
+    * `gen` will make an effort to keep the resulting value's type as narrow as possible, which can be useful for typeclass families. As a
+    * contrived example, when configured with `type Typeclass[x] = Either[Any, x]` and corresponding `join` and `split`, `gen` may derive an
+    * `Either[String, T]` if the target type and available instances allow it. To take advantage of it, one has to define `join` and `split`
+    * in such a way that they propagate narrowed typeclasses as appropriate:
+    *
+    * {{{
+    * def join[L, T](caseClass: CaseClass[Either[L, *], T]): Either[L, T]
+    * }}}
```
nigredo-tori, line 57: This describes the behavior introduced in #628

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -847,6 +852,66 @@ object Magnolia {
...
+        baseClass.asType.toType.decls.collectFirst {
+          case tpe: TypeSymbol if tpe.name == TypeClassNme =>
+            tpe.toType.asSeenFrom(prefixType, baseClass)
+        }
+      }
+      val typeclass = typeDefs.headOption.fold(
+        c.abort(c.enclosingPosition, s"the derivation $prefixObject does not define the Typeclass type constructor")
+      )(_.typeConstructor)
+
+      TypeConstructor(typeclass, appliedType(typeclass, _))
```
nigredo-tori: This was previously done in `gen`.

core/src/main/scala/magnolia1/magnolia.scala
```diff
@@ -847,6 +852,66 @@ object Magnolia {
...
+      import c.internal.polyType
+
+      val tpe = tag.tpe
+
+      def fail() =
+        c.abort(
+          c.enclosingPosition,
+          s"""expected a * -> * HKT,
+             |got: $tpe as ${showRaw(tpe)}
+             |eta-expanded: ${tpe.etaExpand} as ${showRaw(tpe.etaExpand)}"""
```
nigredo-tori, line 887: This is probably too verbose, but I feel like the more information we get here the better. The end users probably wouldn't see this error.

examples/src/main/scala/magnolia1/examples/schema.scala
```diff
@@ -0,0 +1,40 @@
+package magnolia1.examples.schema
+
+import magnolia1._
+
+// Slimmed down version of schema derivation from the Caliban library.
```
nigredo-tori, line 5: Combines code from [here](https://github.com/ghostdogpr/caliban/blob/1c72a10561e2376e5358c36efafb093c858ff99e/core/src/main/scala-2/caliban/schema/SchemaDerivation.scala#L311) and [here](https://github.com/ghostdogpr/caliban/blob/1c72a10561e2376e5358c36efafb093c858ff99e/macros/src/main/scala-2/caliban/schema/DerivedMagnolia.scala#L12-L14). There's some typer weirdness happening there, so I wanted to make sure it actually worked as intended.

test/src/test/scala/magnolia1/tests/tests.scala
```diff
@@ -455,8 +455,8 @@ class Tests extends munit.FunSuite {
         case class Beta(alpha: Alpha)
         Show.gen[Beta]
       """)
-      assert(error contains """
-        |magnolia: could not find Show.Typeclass for type Double
+      assert(clue(error) contains """
+        |magnolia: could not find magnolia1.examples.Show.Typeclass[Double]
```
nigredo-tori, line 459: The error messages were changed to fully show the required type.

examples/src/main/scala/magnolia1/examples/collectFields.scala
```diff
@@ -14,28 +14,44 @@ object CollectFields {
...
+
+  object genDerivation extends Derivation {
+    type Typeclass[A] = CollectFields[Any, A]
+
+    implicit def gen[A]: Typeclass[A] = macro Magnolia.gen[A]
+  }
+
+  object genNarrowDerivation extends Derivation {
+    implicit def genNarrow[Tc[_] <: CollectFields[Any, _], A]: Tc[A] =
+      macro Magnolia.genNarrow[Tc, A]
```
nigredo-tori, line 35: I'd like to be able to write
```scala
implicit def genNarrow[Out, A]: CollectFields[Out, A] =
      macro Magnolia.genNarrow[CollectFields[Out, *], A]
```
However, for whatever reason, this seems to lose the information about what `Out` actually is.

Wrapping the whole thing in another macro (as in the schema example) seems to solve this, but it's a weird magical workaround for what looks like a compiler bug.

adamw:
Thanks, and sorry for the delay :)

PR #645, opened 2026-05-07
fix annotations and inherited annotations not being found for Scala 3 parameterless enum cases

### LINKED ISSUE #491: Annotaions that are attached to the enum elements are not visible
While working on https://github.com/sksamuel/avro4s/pull/791 I realized that the 2nd item in the todo list is blocked by the fact we couldn't obtain annotations that are attached to the enum elements via magnolia (1.3.3).

In particular, I'm testing with

```scala
import com.sksamuel.avro4s.AvroSortPriority

enum ColoursAnnotatedEnum:
  @AvroSortPriority(0)
  case Red
  @AvroSortPriority(1)
  case Amber
  @AvroSortPriority(2)
  case Green
```

Is this a well-known issue that won't be fixed in magnolia?

arnaud-daroussin:
Hi,

I would like to fix #491 to get annotations (and inherited annotations) on Scala 3 parameterless enum cases.

Here is my proposal to fix it. Does it look good to you?

Thanks.

### DIFF
--- a/core/src/main/scala/magnolia1/macro.scala
+++ b/core/src/main/scala/magnolia1/macro.scala
@@ -180,17 +180,18 @@ object Macro:
 
     def anns: Expr[List[scala.annotation.Annotation]] =
       Expr.ofList {
-        tpe.typeSymbol.annotations
+        annotatedSymbol.annotations
           .filter(filterAnnotation)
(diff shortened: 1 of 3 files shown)


adamw:
Thank you!

PR #652, opened 2026-06-16
Fix SealedTrait.Subtype.typeAnnotations to expose subtype's type annotations

### LINKED ISSUE #646: SealedTrait.Subtype.typeAnnotations contains parent's paramTypeAnns instead of subtype's typeAnns
I was deriving a typeclass over a sealed hierarchy and noticed that the `typeAnnotations` exposed on each `SealedTrait.Subtype` did not match what the documentation suggested. For an annotated subtype `@myann case class Cat(...) extends Animal`, `sub.typeAnnotations` for the `Cat` subtype is empty (or contains unrelated parent annotations) instead of `@myann`.

Tracing the macro, `subtypesFromMirrorStep` passes `paramTypeAnns[A]` — the *parent sealed trait's* parameter type annotations — into each subtype's `Subtype` constructor in the field where it should be passing `typeAnns[s]` (the subtype's own type annotations).

Reproducer (scala-cli):

```scala
//> using scala 3.8.3
//> using dep com.softwaremill.magnolia1_3::magnolia:1.3.18

import magnolia1.*

class myann extends StaticAnnotation

sealed trait Animal
case class Dog(name: String) extends Animal
@myann case class Cat(name: String) extends Animal

trait Show[A]:
  def show(a: A): String

object Show extends Derivation[Show]:
  def join[T](cc: CaseClass[Show, T]): Show[T] = (t: T) =>
    cc.params.map(p => s"${p.label}=${p.typeclass.show(p.deref(t))}").mkString(s"${cc.typeInfo.short}(", ",", ")")

  def split[T](st: SealedTrait[Show, T]): Show[T] = (t: T) =>
    st.choose(t) { sub =>
      println(s"sub=${sub.typeInfo.short} typeAnnotations=${sub.typeAnnotations.toList}")
      sub.typeclass.show(sub.value)
    }

  given Show[String] = identity

@main def run(): Unit =
  val show = summon[Show[Animal]]
  show.show(Cat("whiskers"))
  // Expected: sub=Cat typeAnnotations=List(@myann)
  // Actual:   sub=Cat typeAnnotations=List() (or whatever paramTypeAnns[Animal] resolves to)
```

The line:

https://github.com/softwaremill/magnolia/blob/989df299f8f678ae3d0db1c12a8326810832eae0/core/src/main/scala/magnolia1/impl.scala#L286

Compare it with a few lines above where the `anns[s]` and `inheritedAnns[s]` arguments are correctly tied to the subtype:

https://github.com/softwaremill/magnolia/blob/989df299f8f678ae3d0db1c12a8326810832eae0/core/src/main/scala/magnolia1/impl.scala#L284-L286

The fix is changing `IArray.from(paramTypeAnns[A])` to `IArray.from(typeAnns[s])` on line 286. Note: line 255 (`SealedTrait` itself) is the parent trait's slot, so `paramTypeAnns[A]` there is at least defensible — though strictly speaking that one would also be more naturally `typeAnns[A]`. Worth a separate look.

Happy to PR.

adamw:
Fixes #646.

The `SealedTrait.Subtype.typeAnnotations` property currently returns the parent sealed trait's type annotations instead of the individual subtype's own type annotations.

This PR adds a failing test case (`sealed trait enumeration should provide subtype type annotations`) that demonstrates the expected behavior: when accessing `typeAnnotations` on a subtype, it should return that subtype's own type annotations (e.g., `MyTypeAnnotation(2)` for the Attributed subtype in the test) rather than the parent's `paramTypeAnns`.

Implementation of the fix is pending.

Closes softwaremill/magnolia#646.

### DIFF
--- a/test/src/test/scala/magnolia1/tests/AnnotationsTests.scala
+++ b/test/src/test/scala/magnolia1/tests/AnnotationsTests.scala
@@ -80,6 +80,11 @@ class AnnotationsTests extends munit.FunSuite:
     assertEquals(subtypeAnnotations(1).map(_.toString).mkString, "MyAnnotation(0)") // Soccer
   }
 
+  test("sealed trait enumeration should provide subtype type annotations") {
+    val subtypeTypeAnnotations = SubtypeInfo.derived[AttributeParent].subtypeTypeAnnotations
+    assertEquals(subtypeTypeAnnotations.head.map(_.toString).mkString, "MyTypeAnnotation(2)") // Attributed
+  }
+
   test("serialize case class with Java annotations by skipping them") {
     val res = Show.derived[MyDto].show(MyDto("foo", 42))
     assertEquals(res, "MyDto{MyAnnotation(0)}(foo=foo,bar=42)")


adamw:
Here is the failure summary:

---

**CI Failure: `magnolia1.tests.AnnotationsTests` — subtype type annotations test**

The new test added in this PR is failing:

```
==> X magnolia1.tests.AnnotationsTests.sealed trait enumeration should provide subtype type annotations  0.03s
   munit.ComparisonFailException: AnnotationsTests.scala:85
84:    val subtypeTypeAnnotations = SubtypeInfo.derived[AttributeParent].subtypeTypeAnnotations
85:    assertEquals(subtypeTypeAnnotations.head.map(_.toString).mkString, "MyTypeAnnotation(2)") // Attributed
      Obtained empty output!
   => Expected: MyTypeAnnotation(2)
```

**Root cause:** `subtypeTypeAnnotations` returns an empty list for the `Attributed` subtype, even though `@MyTypeAnnotation(2)` is annotated on it. The fix to `SealedTrait.Subtype.typeAnnotations` (which this PR is meant to provide) has not yet been implemented — the failing test is the one added to drive that fix.

The test runs on all three targets (`test3`, `testJS3`, `testNative3`) and fails identically on all of them.

PR #654, opened 2026-06-16
Make Monadic[Future].point use Future.successful

### LINKED ISSUE #647: Monadic[Future].point uses Future(value) instead of Future.successful(value)
The `Monadic[Future]` instance in `magnolia1/monadic.scala` defines `point` as `Future(value)`. By-name evaluation submits a task to the `ExecutionContext` even though `point` is meant to wrap an already-evaluated value. `Future.successful(value)` is the eager, allocation-only counterpart and is what the other instances do (`Some(value)`, `List(value)`).

The practical fallout: every `point` allocates a runnable and bounces through the thread pool. If the EC is shut down or saturated the resulting `Future` is delayed or fails outright. It also breaks the spirit of `pure`/`point` — the value should be lifted, not enqueued.

Reproducer (scala-cli):

```scala
//> using scala 3.8.3

import scala.concurrent.{Future, ExecutionContext}

@main def repro(): Unit =
  given ExecutionContext = ExecutionContext.global

  // What magnolia's Monadic[Future].point currently does:
  val f1 = Future(100)
  // What it should do:
  val f2 = Future.successful(100)

  println(s"Future(100).isCompleted right after creation:           ${f1.isCompleted}")
  println(s"Future.successful(100).isCompleted right after creation: ${f2.isCompleted}")
```

Output (a typical run):

```
Future(100).isCompleted right after creation:           false
Future.successful(100).isCompleted right after creation: true
```

The line:

https://github.com/softwaremill/magnolia/blob/989df299f8f678ae3d0db1c12a8326810832eae0/core/src/main/scala/magnolia1/monadic.scala#L26

Compare with the `Option` and `List` instances directly above, which are eagerly constructed and don't take an `ExecutionContext`:

https://github.com/softwaremill/magnolia/blob/989df299f8f678ae3d0db1c12a8326810832eae0/core/src/main/scala/magnolia1/monadic.scala#L13-L20

One-line fix:

```scala
def point[A](value: A): Future[A] = Future.successful(value)
```

Happy to PR.

adamw:
Fixes #647

The `point` method in `Monadic[Future]` was using `Future(value)` to lift values, which unnecessarily submits the wrapping operation as a task to the ExecutionContext. This can be problematic if the EC is unavailable or constrained.

Changed to `Future.successful(value)`, which returns an already-completed Future without requiring ExecutionContext execution. This is the idiomatic pattern for lifting already-computed values into the Future monad.

Added a test that verifies the behavior using a no-op ExecutionContext—the old implementation would never complete while the new one completes immediately.

Closes softwaremill/magnolia#647.

### DIFF
--- a/core/src/main/scala/magnolia1/monadic.scala
+++ b/core/src/main/scala/magnolia1/monadic.scala
@@ -23,7 +23,7 @@ object Monadic:
       from.flatMap(fn)
 
   given (using ec: ExecutionContext): Monadic[Future] with
-    def point[A](value: A): Future[A] = Future(value)
+    def point[A](value: A): Future[A] = Future.successful(value)
     def map[A, B](from: Future[A])(fn: A => B): Future[B] = from.map(fn)
     def flatMap[A, B](from: Future[A])(fn: A => Future[B]): Future[B] =
       from.flatMap(fn)
--- a/test/src/test/scala/magnolia1/tests/MonadicTests.scala
+++ b/test/src/test/scala/magnolia1/tests/MonadicTests.scala
@@ -1,0 +1,20 @@
+package magnolia1.tests
+
+import magnolia1.*
+import scala.concurrent.{ExecutionContext, Future}
+
+class MonadicTests extends munit.FunSuite:
+
+  test("Monadic[Future].point lifts the value eagerly without using the ExecutionContext") {
+    // An ExecutionContext that never runs any submitted task. With this EC,
+    // Future.successful(value) is already completed, while Future(value)
+    // (which submits a task) never completes.
+    given ExecutionContext = new ExecutionContext:
+      def execute(runnable: Runnable): Unit = ()
+      def reportFailure(cause: Throwable): Unit = ()
+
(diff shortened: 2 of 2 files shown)


adamw:
Here is the failure summary:

---

**CI failure: `MonadicTests` — `point` does not use `Future.successful`**

One test failed across all three platforms (`test3`, `testJS3`, `testNative3`):

```
==> X magnolia1.tests.MonadicTests.Monadic[Future].point lifts the value eagerly without using the ExecutionContext
munit.FailException: MonadicTests.scala:18
  assert(f.isCompleted, "point(value) should produce an already-completed Future")
```

The test asserts that `Monadic[Future].point(value)` returns an already-completed `Future` (i.e. it should use `Future.successful` internally), but the assertion `f.isCompleted` fails — meaning the current implementation wraps the value in `Future.apply` (which dispatches to an `ExecutionContext`), so the future is not immediately completed.

**Fix needed:** Change the `Monadic[Future]` `point` implementation to use `Future.successful(value)` instead of `Future(value)`.

PR #655, opened 2026-06-16
Fix/scala3 covariant recursive derivation

### LINKED ISSUE #424: Recursive derivation not working
probably the same as #401 (not sure)

While working on some examples with @guihardbastien, we can see with this code when the recursion fails: https://scastie.scala-lang.org/ahoy-jon/ijmFz87ESKO2vr6Syf3qRw/13

```scala
import language.experimental.macros, magnolia1._


trait Tc[T] 

trait TcObj[T] extends Tc[T]


object ShowDerivation {
  type Typeclass[T] = Tc[T]
  
  
  def join[T](ctx: CaseClass[Tc, T]): TcObj[T] = new TcObj[T] {}

  def split[T](ctx: SealedTrait[Tc, T]): TcObj[T] = new TcObj[T] {}
  

  def gen[T]: Tc[T] = macro Magnolia.gen[T]
  
  def genObj[T]: TcObj[T] = macro Magnolia.gen[T]
}


sealed trait A

case object B extends A

case class C(next:A) extends A


ShowDerivation.gen[B.type] //success

ShowDerivation.genObj[B.type] //success

ShowDerivation.gen[A] //success

//ShowDerivation.genObj[A] //failure

/*
type mismatch;
 found   : Playground.ShowDerivation.Typeclass[Playground.A]
    (which expands to)  Playground.Tc[Playground.A]
 required: Playground.TcObj[Playground.A]
*/


implicit val tca:Tc[A] = null

ShowDerivation.gen[C] //success

ShowDerivation.genObj[C] //success
```

Arolg:
fix [424](https://github.com/softwaremill/magnolia/issues/424)

### DIFF
--- a/core/src/main/scala/magnolia1/magnolia.scala
+++ b/core/src/main/scala/magnolia1/magnolia.scala
@@ -16,7 +16,7 @@ trait CommonDerivation[TypeClass[_]]:
     */
   def join[T](caseClass: CaseClass[Typeclass, T]): Typeclass[T]
 
-  inline def derivedMirrorProduct[A](
+  transparent inline def derivedMirrorProduct[A](
       product: Mirror.ProductOf[A]
   ): Typeclass[A] = join(CaseClassDerivation.fromMirror(product))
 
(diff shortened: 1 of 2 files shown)


adamw:
The CI fails on formatting.

Could you explain the fix? Also, didn't `transparent` have negative performance impact on the macro expansion? Or on type inference in IDEs?

bookbigwittaya21-web:
BIGWIN×200%

PR #662, opened 2026-07-07
Support for type aliases

OlegYch:
This dealiases TypeRepr to preserve default values and correctly render type names

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,0 +1,2 @@
+-J-Xmx4G
+-J-Xss16M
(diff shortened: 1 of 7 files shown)


adamw:
Thank you :)

ghostdogpr:
This broke a particular use case in Caliban, where we reads annotations from union type aliases. Is it possible to collect and preserve annotations on the alias + combine with those on the dealiased type?

ghostdogpr:
Reference: https://github.com/ghostdogpr/caliban/pull/3058

adamw:
Done :)

PR #665, opened 2026-07-14
Preserve annotations defined on type aliases

adamw:
## Problem

#662 (released in 1.3.22) dealiases the `TypeRepr` before collecting annotations, which dropped annotations placed on the type alias itself. This broke Caliban's annotated union type aliases (see the [failed build](https://github.com/ghostdogpr/caliban/pull/3058) and [this comment](https://github.com/softwaremill/magnolia/pull/662#issuecomment-4964831050)):

```scala
@GQLName("Payload2")
@GQLDescription("Union type Payload")
type Payload = Foo | Bar | Baz
```

`Macro.anns[Payload]` returned `Nil` on 1.3.22, whereas 1.3.21 returned the alias's annotations.

## Fix

`CollectAnnotations.anns` now collects the annotations of the pre-dealias alias symbol and combines them with those of the dealiased type (alias annotations first). For non-alias types the symbols are identical, so behavior is unchanged; opaque types are unaffected since `.dealias` doesn't unwrap them.

Notes on scope:
- Chained aliases (`type A2 = A1; type A1 = B` with annotations on `A1`) only pick up the outermost alias's annotations — there's no one-step dealias in the reflection API.
- Only `anns` is extended; `inheritedAnns`/`typeAnns`/`paramAnns` are left as-is.

## Tests

Two new tests in `TypeAliasesTests`: one reproducing the Caliban break (annotated union type alias), one checking that alias annotations are combined with those on the dealiased type. Both fail without the fix; full suite passes with it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/main/scala/magnolia1/macro.scala
+++ b/core/src/main/scala/magnolia1/macro.scala
@@ -184,10 +184,15 @@ object Macro:
 
     def anns: Expr[List[scala.annotation.Annotation]] =
       Expr.ofList {
-        annotatedSymbol.annotations
+        (aliasAnnotations ++ annotatedSymbol.annotations)
           .filter(filterAnnotation)
           .map(_.asExpr.asInstanceOf[Expr[scala.annotation.Annotation]])
       }
+
+    // annotations on a type alias itself, e.g. @ann type T = ...
+    private def aliasAnnotations: List[Term] =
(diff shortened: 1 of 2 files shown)


ghostdogpr:
Thanks!

PR #666, opened 2026-07-14
more stable tests

OlegYch:
ensure tests don't depend on terminal settings
remove workaround for https://github.com/scalameta/munit/issues/1119

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -74,6 +74,10 @@ lazy val test = (projectMatrix in file("test"))
     projectDependencies ++= Seq(
       "org.scalameta" %%% "munit" % "1.0.0-M12"
     ),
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -74,6 +74,10 @@ lazy val test = (projectMatrix in file("test"))
     projectDependencies ++= Seq(
       "org.scalameta" %%% "munit" % "1.0.0-M12"
     ),
+    scalacOptions ++= List(
+      "-pagewidth",
```
adamw, line 78: why is that option needed and what does it do?

OlegYch: it ensures that compiler wraps error messages consistently, and doesnt depend on current terminal settings

